PR #18, opened 2018-11-27
Doc 4038

RichardSmedley:


modules/ROOT/pages/managing-connections.adoc
```diff
@@ -44,6 +44,15 @@ CouchbaseCluster cluster = CouchbaseCluster.fromConnectionString(connectionStrin
 WARNING: When creating a cluster from a connection string, the Java SDK ignores the scheme component; if a secure connection is desired, the Couchbase environment must be configured as described in <<ssl>>.
 The Java SDK also ignores any options in the connection string, and will not work with a connection string that specifies a bucket name.
 
+[NOTE]
+====
+From Couchbase Java SDK 2.6.0, [.api]`AlternateAddress` is implemented, for connecting to nodes in a NATed environment, such as Docker containers. It is on by default: an xref:client-settings.adoc[environmental setting] can be made, or the SDK will simply compare the bootstrap list, and make the right choice for addressing the alternate address in a multi-network configuration.
```
ingenthr: "Docker containers" is not really a complete description.  It's a nitpick, but it's really with Docker portmapping as Docker Swarm wouldn't have a problem.

Also, I would make mention that it's on by default in the SDK "if the server provides a topology" that includes a multi-network configuration.  

Finally, it's worth mentioning when you might want to disable it.  You'll want to disable it when you're running the app using the SDK "inside" the NAT, such as with Docker Swarm or inside the virtual network at a cloud provider.

modules/ROOT/pages/start-using-sdk.adoc
```diff
@@ -117,6 +117,8 @@ Cluster cluster = CouchbaseCluster.create();
 Cluster cluster = CouchbaseCluster.create("10.0.0.1", "10.0.0.2");
 ----
 
+NOTE: By default, from Couchbase Java SDK 2.6.0, [.a[i]`AlternateAddress` is implemented, for connecting to nodes in a NATed environment, such as Docker containers. An xref:client-settings.adoc[environmental setting] can be made, or the SDK will simply compare the bootstrap list, and make the right choice for addressing the alternate address in a multi-network configuration.
```
ingenthr: I don't think I'd put this much in the "start using".  What do you think?  I think it might be better to just say "if you have an environment with port mapping, see multi-network configuration under managing connections." or some such.

ingenthr (changes requested):

modules/ROOT/pages/start-using-sdk.adoc
RichardSmedley: Yes, I thought I'd taken this out before I did git push.
Thanks for the alternate suggestion :+1:

ingenthr (approved):

PR #19, opened 2018-11-29
DOC-4491 / DOC-4492

RichardSmedley:
FLE - dependency note & example link


simon-dew (approved):
I think the explanation in line 12 could be clearer — does it mean "... the SDK does not have a dependency upon a *third-party* encryption library"?

PR #24, opened 2019-01-03
Update release notes about JDK compatibility

bsubhashni:


modules/ROOT/pages/relnotes-java-sdk.adoc
```diff
@@ -5,7 +5,7 @@
 
 At least Java 7 is required for current releases; all newer, vendor supported Java versions are supported (and encouraged). 
 See the xref:compatibility-versions-features.adoc[Compatibility] section for details.
-Java 8 or later is recommended since it brings performance enhancements and lambda expressions which are very convenient when using the asynchronous API.
+Java 11 is recommended which has various enhancements like lambda local variable type inference, profiling tools and updated security features.
```
ingenthr: I like recommending this, but I think we need to be clear that it's 2.7.2 and later.  I know there aren't likely issues with earlier versions, but we didn't test with Java 11 until 2.7.2.

Maybe just add "(starting with 2.7.2)" after "is recommended"?

I know it's annoying, but I kinda want to be able to be clear that say SDK 2.4 isn't supported on Java 11.

modules/ROOT/pages/relnotes-java-sdk.adoc
```diff
@@ -28,7 +28,7 @@ The other dependency is RxJava which is both used internally and exposed as a fi
 
 == Version 2.7.2 (5 December 2018)
 
-Version 2.7.2 is the third release of the 2.7 series, bringing new features, enhancements, and bugfixes over the last stable release.
+Version 2.7.2 is the third release of the 2.7 series, bringing new features, enhancements, and bugfixes over the last stable release. As of this release, OpenJDK 11 is fully supported. See the xref:compatibility-versions-features.adoc[Compatibility] section for details.
```
ingenthr: Both OpenJDK 11 and Oracle JDK 11.

Also, you should mention with the HotSpot JVM.

I'd remove the word "fully", as it implies there is some other partial support otherwise.

ingenthr (changes requested):

modules/ROOT/pages/relnotes-java-sdk.adoc
bsubhashni: Sure, fixed.

modules/ROOT/pages/relnotes-java-sdk.adoc
bsubhashni: Sure, fixed.

ingenthr (approved):

RichardSmedley (approved):
LGTM :+1:

PR #38, opened 2019-07-08
updated After 'is' has to After 'it' has

ajay143444:
there was a grammatical mistake in document, updated After 'is' has extracted the rows to After 'it' has extracted the rows at line 392 to correct the same


RichardSmedley (approved):
LGTM
Thanks @ajay143444 :+1:

PR #41, opened 2019-07-23
the line 66 was confusing to end user

ajay143444:
the line 66 is confusing to readers so updated it to make it more clear.


RichardSmedley (approved):
LGTM Thanks

PR #44, opened 2019-07-30
Updated the doc to give a better understanding

ajay143444:
Updated the doc to give a better understanding.


RichardSmedley:
Thanks for the PR, but this doesn't really clarify.
I've changed the text to read each, though, which I hope you find clearer.

PR #45, opened 2019-07-30
provided link for MapReduce

ajay143444:
Added link for MapReduce documentation at line 146 so that user can go to it directly to Map Reduce doc from here. I have added the hyper link in doc, please review it.
Below is the URL which it should open.
https://docs.couchbase.com/java-sdk/2.7/view-queries-with-sdk.html


RichardSmedley (approved):
LGTM

PR #57, opened 2019-08-16
Corrected line 103

ajay143444:
It looks like that the correct thing will be ClusterEnvironment will get auto shut down instead but its written also shutdown. Please have a look.


RichardSmedley (approved):
LGTM

PR #72, opened 2019-10-10
DOC-5842 link GH examples at top of doc

RichardSmedley:


modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -18,6 +18,11 @@ This is the beta release of Distributed Transactions for Couchbase, providing th
 
 Please see xref:6.5@server:learn:data/distributed-acid-transactions.adoc[our introduction to ACID Transactions] for a guide to the benefits and trade-offs of multi-document transactions.
 
+Below we show you how to create Transactions, step-by-step, and what happens when something goes wrong.
+You may also want to start with our https://github.com/couchbaselabs/couchbase-transactions-java-examples[GitHub transactions examples page], 
+which features two useful downloadable examples of using Distributed Transactions.
```
programmatix: I'd remove the "two", it's likely to grow over time.

Would also remove "and what happens when something goes wrong", it's just a negative start to the doc.

programmatix (changes requested):

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
RichardSmedley: Sounds reasonable

programmatix (approved):

PR #78, opened 2019-12-06
WIP sdk3 client settings

dnault:
Updated the overview section for SDK3's new way of configuring the cluster environment.

Updated the timeout config section.

### DIFF
--- a/modules/ref/pages/client-settings.adoc
+++ b/modules/ref/pages/client-settings.adoc
@@ -295,6 +295,14 @@ System Property: `com.couchbase.env.io.idleHttpConnectionTimeout`
 +
 The length of time an HTTP connection may remain idle before it is closed and removed from the pool.
 
+Name: *Config Poll Interval*::
+Builder Method: `IoConfig.configPollInterval(Duration)`
++
+Default:  `2.5s`
++
+System Property: `com.couchbase.env.io.configPollInterval`
++
+The interval at which the client fetches cluster topology information in order to proactively detect changes.
 
 === Circuit Breaker Options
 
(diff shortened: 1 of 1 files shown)


RichardSmedley (commented):
Looking good - thanks David :+1: 
I've tweaked one thing (callouts within the ---- listing need to be comments //).
.
btw undocumented environment settings? (line 99) - do we have many of these (or will we, once we've finished this doc)?
.
also, having removed type information for Duration (no longer a long) in the Timeout Options section (line 256, onwards), 
would it be a good idea to repeat the info from lines 92 & 93 about what goes there? (in case someone follows a link to the anchor, instead of the top of the page)
.
Anyway, great to see these things being well explained. Many thanks.

dnault:
> btw undocumented environment settings? (line 99) - do we have many of these (or will we, once we've finished this doc)?

Heresy, I know! This TIP is perhaps not so useful to end-users. Let's delete it.

> also, having removed type information for Duration (no longer a long) in the Timeout Options section (line 256, onwards),

The type information is still there... we're just using java.util.Duration ("Duration" for short) in SDK 3 instead of longs.

> would it be a good idea to repeat the info from lines 92 & 93 about what goes there? (in case someone follows a link to the anchor, instead of the top of the page)

Yes, this would be good. Maybe with a link to the explanation instead of repeating it for every property?

RichardSmedley:
> Yes, this would be good. Maybe with a link to the explanation instead of repeating it for every property?

Good idea :+1:

RichardSmedley:
> > Yes, this would be good. Maybe with a link to the explanation instead of repeating it for every property?
> 
> Good idea

The following can be added to each section, where appropriate:

`See <<duration-values, setting duration values>> under xref:#system-properties[System Properties].`

RichardSmedley (approved):
Looking ready to merge as β doc. :+1:

PR #79, opened 2019-12-19
Fix idleHttpConnectionTimeout default

dnault:
Default for idleHttpConnectionTimeout should be 30s. Also added a note that the server may unilaterally enforce a 1 minute idle timeout.


dnault:
Oh gosh, this PR was based on the wrong brach ::facepalm::

RichardSmedley:
:man_shrugging: We've all done it!

PR #80, opened 2019-12-19
Correct default for idleHttpConnectionTimeout is 30s

dnault:
Default for idleHttpConnectionTimeout is 30s. Also added a note that the server may unilaterally enforce a 1 minute idle timeout.


RichardSmedley (approved):
LGTM
Is it worth also linking to the Health Check concept page, with a mention of using Ping as an alternative keepalive mechanism?

dnault:
> Is it worth also linking to the Health Check concept page

Sure, they do seem to be related.

> with a mention of using Ping as an alternative keepalive mechanism?

I'll have to look more closely, but I'm not sure the Ping command would actually refresh *all* of the connections in the pool. You could Ping all day long, but it might use the same connection for every ping request.

PR #84, opened 2020-01-08
DOC-5878 REQUEST_PLUS example

RichardSmedley:
& DOC-4824

### DIFF
--- a/modules/ROOT/pages/scan-consistency-examples.adoc
+++ b/modules/ROOT/pages/scan-consistency-examples.adoc
@@ -22,7 +22,9 @@ At the other end of the scale there is "request plus" which will — when a quer
 This will make sure that you read your own writes, but it might take some time if there is a write-heavy workload.
 Also sometimes you are only interested in the last mutation(s) you performed, so waiting for everything else in the index to be updated can mean unnecessary latency.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/scan-consistency-examples.adoc
```diff
@@ -22,7 +22,9 @@ At the other end of the scale there is "request plus" which will — when a quer
 This will make sure that you read your own writes, but it might take some time if there is a write-heavy workload.
 Also sometimes you are only interested in the last mutation(s) you performed, so waiting for everything else in the index to be updated can mean unnecessary latency.
 
-This is where "statement plus" comes in, also named "at plus".
+== Read Your Own Writes
+
+This is where "statement plus" comes in: roughly equivalent to building a query with "at_plus" in the other SDKs, this enum also provides the same _Read Your Own Writes (RYOW)_ consistency.
```
daschl: actually this isn't really true.. right now statement_plus and request_plus are the same from a server perspective. I think it was meant to be used if you have multiple statements in a single request? I think the docs should say that it is functionally equivalent to request_plus and is actually removed (!) in sdk3.

daschl (changes requested):

daschl (approved):

PR #89, opened 2020-01-19
Various typo and rephrasing fixes.

ingenthr:
I've aligned these with .NET as going through.  Mostly simplifying, but one typo fix.


RichardSmedley (approved):
:+1:

PR #111, opened 2020-03-11
updated line 25 'an' to 'a'

ajay143444:
the sentence should have 'a' instead of 'an' before query as it is not starting from vowel


AV25242 (approved):

RichardSmedley (approved):
LGTM

PR #114, opened 2020-03-11
correcting line 90

ajay143444:
there is some error in line 90 which was making it confusing. I have tried to correct it to improve the meaning


AV25242 (approved):

AV25242 (approved):

RichardSmedley:
Thanks.
_On the one hand ... but on the other_ is a colloquialism,
as it's making the doc harder to understand, I have removed it: 300349d

PR #122, opened 2020-04-27
Updated Examples with Healthcheck

ingenthr:
I have removed each commit individually so I could update the dependencies in the project, then added a healthcheck example based on the work of @RichardSmedley.

The code sample shows both ping and diagnostics.

@RichardSmedley: can you update this to fold into the iteration starting April 27?


RichardSmedley (approved):
LGTM :+1:

PR #134, opened 2020-06-05
WaitUntilReady()

RichardSmedley:
DOC-6912

### DIFF
--- a/modules/howtos/pages/managing-connections.adoc
+++ b/modules/howtos/pages/managing-connections.adoc
@@ -230,6 +230,41 @@ INFO: DNS SRV enabled, but less or more than one seed node given.
 Proceeding with normal bootstrap.
 ----
 
+
+== Waiting for Bootstrap Completion
+
+Sometimes, cluster resources are not ready, and a simple connection attempt will result in a timeout error.
+In these circumstances, wrapping the first call to the Cluster object in waitUntilReady() will avoid such an error.
+
+Add to the `cluster()` in the xref:#connecting-to-a-cluster[earlier example]:
+
+[source,java]
+----
+Cluster cluster = Cluster.connect("127.0.0.1", "Administrator", "password");
+cluster.waitUntilReady(Duration.ofSeconds(10));
+Bucket bucket = cluster.bucket("travel-sample");
+Collection collection = bucket.defaultCollection();
+----
+
+Or more fully:
+
+[source,java]
+----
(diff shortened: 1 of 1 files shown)


modules/howtos/pages/managing-connections.adoc
```diff
@@ -230,6 +230,41 @@ INFO: DNS SRV enabled, but less or more than one seed node given.
 Proceeding with normal bootstrap.
 ----
 
+
+== Waiting for Bootstrap Completion
+
+Sometimes, cluster resources are not ready, and a simple connection attempt will result in a timeout error.
```
daschl: Maybe this can be made a little more clear.. something like "Depending on the environment and network latency, bootstrapping the SDK fully might take a little longer. Since the default Key/Value timeout is only 2.5 seconds, you might see timeouts during bootstrap. To prevent those early timeouts from happening, you can use the waitUntilReady method .... or something like this?

modules/howtos/pages/managing-connections.adoc
```diff
@@ -230,6 +230,41 @@ INFO: DNS SRV enabled, but less or more than one seed node given.
...
+In these circumstances, wrapping the first call to the Cluster object in waitUntilReady() will avoid such an error.
```
daschl: the call is not wrapped?

modules/howtos/pages/managing-connections.adoc
```diff
@@ -230,6 +230,41 @@ INFO: DNS SRV enabled, but less or more than one seed node given.
...
+    Cluster cluster = Cluster.connect("127.0.0.1", "Administrator", "password");
+    cluster.waitUntilReady(Duration.ofSeconds(10));
+    Bucket bucket = cluster.bucket("travel-sample");
+    Collection collection = bucket.defaultCollection();
+  }
+}
+----
+
+// Other timeout issues may occur when using the SDK located geographically separately from the Couchbase Server cluster -- this is xref:project-docs:compatibility#[not recommended].
+// See the xref:#working-in-the-cloud[Cloud section] below for some suggestions of settings adjustments.
```
daschl, line 267: Also, we need to point out that at the bucket level there is also a waitUntilReady(). the difference is that at the bucket level it also takes bucket-level KV into account (in addition to all of the cluster-level wait until ready)

daschl (approved):

PR #136, opened 2020-06-23
1.0.1 release.

programmatix:
Note that the documentation really needs updating to reflect the new additions to the API.
I have this work underway but it's turned into a larger rework than anticipated.  I don't want to block the 1.0.1 release any longer, so those changes will be delivered separately.


RichardSmedley (approved):
:+1:

PR #137, opened 2020-06-23
JCBC-1636: Update concept docs to Java SDK 3

daschl:
These were idntified as outstanding, so polish and update.


RichardSmedley (approved):
LGTM

PR #138, opened 2020-06-23
improve get() example

steveyen:


ingenthr (changes requested):
I agree with where you're going conceptually, but the change was to the "full sample" without changing the linked code itself.

For instance on this line you'll see a link to a file:
https://github.com/couchbase/docs-sdk-java/pull/138/files#diff-52933e985f04c31583c642692cb8c13eR91

That links to, in this case, this file:
https://github.com/couchbase/docs-sdk-java/blob/release/3.0/modules/hello-world/examples/StartUsing.java

A more complete change would be updating the code snippet and the full example. 

Note that I think in all of the SDK docs, the only place we have (intentionally) a full example is here.  In the future, @amarantha-k and team are going to have a way to link to the full code example from the tagged samples in the page.  In all of the SDK cases, the complete, compilable (and frequently in a project) code examples should be there.

ingenthr:
Related here, @RichardSmedley: the goal is for these to be common across the SDKs based on switching, right?  Would you be able to add the other changes?

RichardSmedley (approved):
Thanks @steveyen 
.
@ingenthr putting through for now, will address the technical debt issues later. :+1:

RichardSmedley:
Added change to https://github.com/couchbase/docs-sdk-java/blob/release/3.0/modules/hello-world/examples/StartUsing.java

PR #147, opened 2020-07-10
Example of document expiration > 30 days

RichardSmedley:
DOC-6968


dnault (commented):
I do not think we should add this to the documentation.

We're fixing the duration bug in 3.0.6, so there's no need for users to do gymnastics to set a duration longer than 30 days.
The user can just call `.expiry(Duration.ofDays(60))`

RichardSmedley:
Looks like the SDK will be fixed to purely take an offset in the next release.
Closing this, and adding an admonition to the doc to explain issue.

PR #160, opened 2020-08-11
xattr example from Roi

RichardSmedley:
DOC-7187 / DOC-7188

### DIFF
--- a/modules/howtos/pages/subdocument-operations.adoc
+++ b/modules/howtos/pages/subdocument-operations.adoc
@@ -396,3 +396,19 @@ NOTE: Currently, paths cannot exceed 1024 characters, and cannot be more than 32
 
 
 include::6.5@sdk:shared:partial$sdk-xattr-overview.adoc[tag=extended_attributes]
+
+
+[source,java]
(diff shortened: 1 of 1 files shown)


modules/howtos/pages/subdocument-operations.adoc
```diff
@@ -396,3 +396,19 @@ NOTE: Currently, paths cannot exceed 1024 characters, and cannot be more than 32
...
+
+
+[source,java]
+.Full Document Example
+----
+    JsonObject docContent = JsonObject.create().put("body", "value");
+collection.mutateIn("doc-id", Arrays.asList(
+            MutateInSpec.upsert("foo", "bar").xattr().createPath(),
+            MutateInSpec.replace("", docContent)),
+            MutateInOptions.mutateInOptions().storeSemantics(StoreSemantics.UPSERT)
```
daschl: this line can be removed to make it simpler, it is not required (don't forget to remove the trailing , on the line before then)

daschl (approved):
+2 still 👍

PR #167, opened 2020-09-04
JCBC-1633 - convert all doc code samples to runnable projects

mikereiche:
Change-Id: I313c6cca9fb14d4cad030279526f04ab31286431

### DIFF
--- a/modules/concept-docs/examples/HealthCheckConcepts.java
+++ b/modules/concept-docs/examples/HealthCheckConcepts.java
@@ -30,7 +30,7 @@ import java.util.Map;
 
 import static com.couchbase.client.java.diagnostics.PingOptions.pingOptions;
 
-public class HealthCheck {
+public class HealthCheckConcepts {
 
   public static void main(String... args) {
     Cluster cluster = Cluster.connect("127.0.0.1", "Administrator", "password");
@@ -41,53 +41,53 @@ public class HealthCheck {
 
     {
 // #tag::ping-basic[]
-PingResult pingResult = cluster.ping();
+      PingResult pingResult = cluster.ping();
 
-for (Map.Entry<ServiceType, List<EndpointPingReport>> service : pingResult.endpoints().entrySet()) {
-  for (EndpointPingReport er : service.getValue()) {
-    System.err.println(
-      service.getKey() + ": " + er.remote() + " took " + er.latency()
-    );
-  }
-}
+      for (Map.Entry<ServiceType, List<EndpointPingReport>> service : pingResult.endpoints().entrySet()) {
+        for (EndpointPingReport er : service.getValue()) {
+          System.err.println(
+              service.getKey() + ": " + er.remote() + " took " + er.latency()
+          );
+        }
+      }
 // #end::ping-basic[]
     }
 
     {
 // #tag::ping-json-export[]
-PingResult pingResult = cluster.ping();
+      PingResult pingResult = cluster.ping();
 
-System.out.println(pingResult.exportToJson());
+      System.out.println(pingResult.exportToJson());
 // #end::ping-json-export[]
     }
 
     {
 // #tag::ping-options[]
-PingResult pingResult = cluster.ping(pingOptions().serviceTypes(EnumSet.of(ServiceType.QUERY)));
+      PingResult pingResult = cluster.ping(pingOptions().serviceTypes(EnumSet.of(ServiceType.QUERY)));
 
-System.out.println(pingResult.exportToJson());
+      System.out.println(pingResult.exportToJson());
 // #end::ping-options[]
     }
 
     {
 // #tag::diagnostics-basic[]
-DiagnosticsResult diagnosticsResult = cluster.diagnostics();
+      DiagnosticsResult diagnosticsResult = cluster.diagnostics();
 
-for (Map.Entry<ServiceType, List<EndpointDiagnostics>> service : diagnosticsResult.endpoints().entrySet()) {
-  for (EndpointDiagnostics ed : service.getValue()) {
-    System.err.println(
-      service.getKey() + ": " + ed.remote() + " last activity  " + ed.lastActivity()
-    );
-  }
-}
+      for (Map.Entry<ServiceType, List<EndpointDiagnostics>> service : diagnosticsResult.endpoints().entrySet()) {
+        for (EndpointDiagnostics ed : service.getValue()) {
+          System.err.println(
+              service.getKey() + ": " + ed.remote() + " last activity  " + ed.lastActivity()
+          );
+        }
+      }
 // #end::diagnostics-basic[]
     }
 
     {
 // #tag::diagnostics-options[]
-DiagnosticsResult diagnosticsResult = cluster.diagnostics();
+      DiagnosticsResult diagnosticsResult = cluster.diagnostics();
 
-System.out.println(diagnosticsResult.exportToJson());
+      System.out.println(diagnosticsResult.exportToJson());
 // #end::diagnostics-options[]
     }
   }
--- a/modules/hello-world/examples/DocParser.java
+++ b/modules/hello-world/examples/DocParser.java
@@ -1,0 +1,368 @@
+/*
+ * Copyright (c) 2020 Couchbase, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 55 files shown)


modules/concept-docs/examples/HealthCheckConcepts.java
```diff
@@ -0,0 +1,100 @@
...
+import com.couchbase.client.java.Scope;
+import com.couchbase.client.java.diagnostics.PingOptions;
+
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Map;
+
+import static com.couchbase.client.java.diagnostics.PingOptions.pingOptions;
+
+public class HealthCheckConcepts {
```
mikereiche, line 33: This is a copy of the HealthCheck class in 'howtos'

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,335 @@
+
```
mikereiche, line 16: This is the class I used to automatically extract the [source,java] sections.

mikereiche (commented):
sample-application.adoc and working-with-collections.adoc remain to be done.

modules/concept-docs/examples/HealthCheckConcepts.java
```diff
@@ -0,0 +1,79 @@
+import com.couchbase.client.core.diagnostics.DiagnosticsResult;
```
daschl, line 17: I think we should add our license header to each?

modules/concept-docs/examples/buckets_and_clusters.java
```diff
@@ -0,0 +1,46 @@
+import com.couchbase.client.java.Bucket;
+import com.couchbase.client.java.Cluster;
+import com.couchbase.client.java.ClusterOptions;
+import com.couchbase.client.java.Collection;
+import com.couchbase.client.java.Scope;
+import com.couchbase.client.java.env.ClusterEnvironment;
+import com.couchbase.client.java.manager.bucket.BucketManager;
+import com.couchbase.client.java.manager.bucket.BucketSettings;
+
+public class buckets_and_clusters {
```
daschl: can you modify the class and file names to use the uppercase syntax like in the topmost file?

daschl (changes requested):
A couple nits I think it just needs a bit manual cleanup (i.e the class names) ..

daschl:
@RichardSmedley do we have a way of making sure the new linked sample are exactly the on as the old after this change?

RichardSmedley:
@daschl  Could build locally and put side-by-side with existing docs?

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,338 @@
...
+		for (String fileName : args) {
+			try {
+				parser.process(fileName);
+			} catch (IOException e) {
+				System.err.println(e);
+			}
+		}
+	}
+
+	String preamble=
```
ingenthr: stringbuilder maybe?  This isn't so idiomatic java, is it?

ingenthr: I could be wrong, so don't take that as a strong statement.  :)

dnault: String concatenation is fine here. StringBuilder only adds value when the string is assembled using branches or loops.

modules/concept-docs/examples/data_model.java
```diff
@@ -0,0 +1,43 @@
...
+    cluster = Cluster.connect(connectionString,
+        ClusterOptions.clusterOptions(username, password).environment(environment));
+    bucket = cluster.bucket(bucketName);
+    scope = bucket.defaultScope();
+    collection = bucket.defaultCollection();
+  }
+
+  public void data_model_1() throws Exception {
+    // #tag::data_model_1[]
+    Map<String,String> myMap = collection.map("name",String.class);
```
dnault: I see some formatting inconsisences... where does this code come from? Can you simply auto-reformat the code, or are there complications?

modules/concept-docs/examples/n1ql_query.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Collections;
+import java.util.UUID;
+
+public class n1ql_query {
+
+  String connectionString="localhost";
+  String username="Administrator";
+  String password="password";
+  String bucketName="travel-sample";
+  String id = "user::" + UUID.randomUUID().toString();
```
dnault: NIT: When an object is concatenated to a string using `+` the compiler automatically calls `toString()... so this could just be

    String id = "user::" + UUID.randomUUID();

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,338 @@
...
+	int fileNo = 0;
+	BufferedWriter codeWriter;
+
+	public static void main(String[] args) throws IOException {
+		DocParser parser = new DocParser();
+		for (String fileName : args) {
+			try {
+				parser.process(fileName);
+			} catch (IOException e) {
+				System.err.println(e);
```
dnault: Better to just let the exception kill the process? If not, `e.printStackTrace()` will aid debugging more than `System.err.println()`.

modules/hello-world/examples/DocParser.java
dnault: NIT: `final`?

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,338 @@
...
+		Queue<String> adoc = new LinkedList<>();
+		String dirName = fileName.substring(0, fileName.lastIndexOf("/"));
+		String baseName = fileName.substring(fileName.lastIndexOf("/")+1, fileName.lastIndexOf(".adoc"));
+		String javaName = baseName
+				.replace("-", "_")
+				.replace(".", "_");
+		String javaFilename = dirName+"/../examples" + "/" + javaName +".java";
+		File javaFile= new File(javaFilename);
+		if ( javaFile.exists()){
+			System.out.println("java file already exists: "+javaFile.getPath());
```
dnault: Should the exiting file be deleted and recreated?

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,338 @@
...
+			return;
+		} else {
+			System.out.println("java file already exists: "+javaFile.getPath());
```
dnault: Is this message correct? I thought the "file exists" was the other branch.

modules/hello-world/examples/DocParser.java
```diff
@@ -0,0 +1,338 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.Stack;
+
+public class DocParser {
+
+
+	int fileNo = 0;
```
dnault: Please auto-reformat with soft tabs. If you can, try to get into the habit of reformatting prior to each commit. (I forget to do it sometimes too, btw).

modules/hello-world/examples/DocParser.java
dnault: So it's a temporary utility used only for converting the existing code samples in docs to Java classes? After the conversion it will go away?

mikereiche: Yes. I was considering saving it in case someone wanted to pull the source examples out of other doc.

modules/concept-docs/examples/buckets_and_clusters.java
mikereiche: Will do.

modules/hello-world/examples/DocParser.java
dnault: Oh! In that case, sorry for all the nit-picking. If it works, it works! :-)

modules/concept-docs/examples/HealthCheckConcepts.java
mikereiche: I've added the license header to all the *.java in here.

modules/hello-world/examples/DocParser.java
mikereiche: I manually edit the files after they are created (remove unnecessary imports etc). So I only want them generated if they don't already exists. ( I have manually renamed the files to be more java-class-like names, so this isn't an issue).

daschl (approved):
I'm ok with this change, but I'm also realistic that it is too big for a thorough review.

Maybe after we get this in, it would be good if @RichardSmedley gives it a close look and we also go over each page and see which examples look funky by accident.

modules/concept-docs/examples/HealthCheckConcepts.java
RichardSmedley: :+1:

RichardSmedley (approved):

PR #186, opened 2020-11-11
JCBC-1659: Refactor Document Expiry Duration

daschl:
This changeset aligns the documentation for expiry (which right now
I think is only covered in the kv ops page) with the newer developments
around relative and absolute values.


RichardSmedley (approved):
Thanks @daschl - LGTM

PR #198, opened 2020-12-22
only SASL PLAIN is allowed for LDAP

RichardSmedley:
DOC-7493

### DIFF
--- a/modules/howtos/examples/Auth.java
+++ b/modules/howtos/examples/Auth.java
@@ -48,7 +48,7 @@ public class Auth {
         .username("username")
         .password("password")
         // enables only the PLAIN authentication mechanism, used with LDAP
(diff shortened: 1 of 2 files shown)


daschl:
I looked this up again and it is a bit confusing. The change you have here is correct, but the old code was not wrong. As a fact, the new method does exactly that:

```java
    public Builder onlyEnablePlainSaslMechanism() {
      return allowedSaslMechanisms(EnumSet.of(SaslMechanism.PLAIN));
    }
```

I think the confusion came about because there is also a builder method called `enablePlainSaslMechanism` which enables PLAIN in addition to all the others, and this is the one we should not use. The javadocs discuss that subject a bit. 

So I think the change is good, and especially for the 3.1 docs this goes in to we don't need to add the sentence that it only is available in 3.0.9 and later? for the 3.0 branch we can leave as-is since the info is not wrong.

PR #199, opened 2021-01-06
Fix compilation error in TransactionsExample

dnault:
Fix compilation error in TransactionsExample.

Compile the devguide examples.


daschl:
ping @programmatix

programmatix:
Ah, yes.  This example is technically broken at the moment - but I have the 'true fix' pending and ready to go out in the next transactions library release, next month, at which point the example will compile.  The example is how I want things to work, so I believe I'd prefer to keep it as-is, and live with the broken-ness - unless someone pushes back strongly on this.  (Worth noting that query support is at alpha level currently.)

dnault:
Graham's wish is my command :-)

PR #202, opened 2021-01-19
DOC-8030 Field Level Encryption v2 migration limitations and details

dnault:
Update the FLE migration docs to emphasize SDK 2 cannot
read fields encrypted by SDK 3.

Recommend an offline migration, or migration using a blue-green
deployment.

Plus a few minor edits.

### DIFF
--- a/modules/howtos/pages/encrypting-using-sdk.adoc
+++ b/modules/howtos/pages/encrypting-using-sdk.adoc
@@ -72,7 +72,7 @@ Two modes of operation are available:
 
 === Data Binding Example
 
-Sensitive fields of your POJOs can be annotated with `@Encrypted`.
+Sensitive fields of your data classes can be annotated with `@Encrypted`.
 Let's use this class as an example:
 
 [source,java]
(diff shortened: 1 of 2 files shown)


modules/howtos/pages/encrypting-using-sdk.adoc
```diff
@@ -162,8 +162,16 @@ include::example$EncryptingUsingSDK.java[tag=encrypting_using_sdk_9,indent=0]
 [#migration-from-sdk2]
 == Migrating from SDK 2
 
-If you were previously using Field-Level Encryption with Java SDK 2, a few extra configuration steps are required.
+SDK 3 can be configured to read fields encrypted by SDK 2.
+However...
 
+WARNING: SDK 2 cannot read fields encrypted by SDK 3.
```
ingenthr, line 165: I think there's a way for a callout box that you could move this above the paragraph, then drop the however…

(also, ellipses are actually a unicode thing!)

ingenthr (approved):
LGTM, but I might recommend moving that admonition above the paragraph and then remove the "However" with the ellipses.  Also, ellipses is actually a Unicode thing …

mikereiche (approved):

RichardSmedley (approved):
LGTM - thanks @dnault

PR #205, opened 2021-01-21
Expand link text so it looks like a link.

dnault:
Motivation
----------
If the link text consists only of literal monospace
(text enclosed in backticks), the docs styling doesn't
give any indivation it's a hyperlink unless you mouse
over it.

A better fix would be to change the docs site CSS,
but that's beyond my ken.

Modifications
-------------
Expand the link text from "`ClusterEnvironment`"
to "configuring the Java SDK's `ClusterEnvironment`".


dnault:
Here's what it looks like before the fix. Notice there's no visual indication that `ClusterEnvironment` is a link.

before

RichardSmedley (approved):
Good workaround - note UI improvements are planned  :+1:

PR #208, opened 2021-02-04
Remove recommendation to avoid duration > 30 days

dnault:
Since Java SDK 3.0.6, it's safe to specify a duration longer than 30 days.

### DIFF
--- a/modules/howtos/pages/kv-operations.adoc
+++ b/modules/howtos/pages/kv-operations.adoc
@@ -209,11 +209,9 @@ You can set an expiry value from a `Duration` when creating a document:
 include::example$KvOperations.java[tag=expiry-insert]
 ----
 
-We recommend using a relative duration only if the provided value is less than 30 days. 
-The reason is that the server will assume any value larger than that to be an absolute Unix timestamp. 
-The SDK tries its best to coerce it into sane values, but to avoid any unexpected behavior please stick with relative duration for durations of less than 30 days.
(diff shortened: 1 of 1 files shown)


modules/howtos/pages/kv-operations.adoc
```diff
@@ -209,11 +209,7 @@ You can set an expiry value from a `Duration` when creating a document:
 include::example$KvOperations.java[tag=expiry-insert]
 ----
 
-We recommend using a relative duration only if the provided value is less than 30 days. 
-The reason is that the server will assume any value larger than that to be an absolute Unix timestamp. 
-The SDK tries its best to coerce it into sane values, but to avoid any unexpected behavior please stick with relative duration for durations of less than 30 days.
-
-For longer expiry values, use an `Instant` instead:
+Alternatively, you can specify the expiry as an `Instant`:
```
osfameron: Does that mean that the "magic" behaviour above has been removed?
e.g. any numeric value is a duration, and only an `Instant` object represents a point in time?

dnault: Prior to Java SDK 3.0.6, specifying a duration longer than 30 days would cause the document to expire IMMEDIATELY... but that's no longer the case. Your comment reminded me that there is still some weirdness at the 50 *year* threshold. Updated the PR accordingly.

daschl (approved):

PR #210, opened 2021-02-18
DOC-7628 scope/collection examples for Java Analytics

osfameron:
In particular I'm looking for comments on whether I've understood the Scoped dataset concept properly: this approach seems to be confirmed from http://review.couchbase.org/c/couchbase-jvm-clients/+/134947

> This adds analyticsQuery() methods to Collection. *In this implementation*,
> the dataverse name for the collection must be <bucket>.<scope> as that
> is what scope.analyticsQuery() uses for query_context

(Highlight of "In this implementation" is mine - is that an important caveat? Will this behaviour change/be implemented differently by other SDKs?)

NB: I've also made some comments about the Async/Reactive examples not running, and some warnings about thread termination. I think those are strictly speaking out of scope for this PR, but I'm happy to fix if anyone has suggestions!

### DIFF
--- a/modules/howtos/examples/Analytics.java
+++ b/modules/howtos/examples/Analytics.java
@@ -14,6 +14,14 @@
  * limitations under the License.
  */
 
+ /*
+  * You need the following datasets created:
+  *
+  *  CREATE DATASET `airports` ON `travel-sample` where type = "airport";
+  *  CREATE DATASET `huge-dataset` ON `travel-sample`;
+  *
+  */
+
 // tag::imports[]
 import com.couchbase.client.core.error.CouchbaseException;
 import com.couchbase.client.java.Cluster;
@@ -36,6 +44,20 @@ public class Analytics {
 
   static Cluster cluster = Cluster.connect("localhost", "Administrator", "password");
 
+  public static void show(String description, AnalyticsResult result) {
+    System.out.println(description);
+
+    for (JsonObject row : result.rowsAsObject()) {
+      System.out.println("Found row: " + row);
+    }
+
+    System.out.println("Reported execution time: "
+      + result.metaData().metrics().executionTime());
+
+      System.out.println();
+  }
+
+
   public static void main(String... args) {
     {
       // tag::simple[]
@@ -49,6 +71,9 @@ public class Analytics {
 
         System.out.println("Reported execution time: "
           + result.metaData().metrics().executionTime());
+
+        System.out.println();
+
       } catch (CouchbaseException ex) {
         ex.printStackTrace();
       }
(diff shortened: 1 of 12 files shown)


modules/howtos/pages/analytics-using-sdk.adoc
```diff
@@ -193,8 +193,43 @@ This query will stream all rows as they become available form the server. If you
...
+CREATE DATASET `airports-collection` ON `travel-sample`.inventory.airport;
+-- compare with:
+-- CREATE DATASET airports ON `travel-sample` WHERE `type` = "airport";
+----
+
+You run the query in exactly the same way as usual:
+
+[source,java]
+----
+include::../examples/Analytics.java[tag=handle-collection,indent=0]
```
RichardSmedley, line 217: Whilst _::./examples/Analytics.java_ will work,
_::example$Analytics.java_ is more idiomatically Antora, and enables Antora macros to do whatever they do should we happen to make more use of them ;)

RichardSmedley (approved):
LGTM, thanks @osfameron  - but I'll let @daschl comment on the Java before we merge.  :+1:

modules/howtos/examples/Analytics.java
```diff
@@ -122,50 +155,76 @@ public static void main(String... args) {
...
-      result
-        .flatMapMany(ReactiveAnalyticsResult::rowsAsObject)
-        .subscribe(row -> System.out.println("Found row: " + row));
-      // end::simplereactive[]
+    {
+      // tag::handle-scope[]
+      Bucket bucket = cluster.bucket("travel-sample");
+      Scope scope = bucket.scope("inventory");
+      AnalyticsResult result = scope.analyticsQuery(
+        "SELECT airportname, country FROM `airports-collection` WHERE country='France' LIMIT 4");
```
daschl, line 180: actually I'm not sure you need to do it like this way with 7.0 .. I think you can query collections directly with analytics now? (so no need to create a specific dataset for it)

daschl (changes requested):
So I think while this approach works, you should be able to query the collection directly in 7.0, or am I mistaken?

modules/howtos/examples/Analytics.java
osfameron: @daschl I couldn't get that to work either via the SDK or the workbench Analytics query -

e.g.  I get an error like “Cannot find dataset airline in dataverse travel-sample.inventory nor an alias with name airline!” if I try to run `select * from travel-sample.inventory.airline` (e.g. against a collection directly, instead of an analytics dataset)

So I'm very Confused! 😕 

Do you have a working simpler example, or other advice?

osfameron:
This is subsumed by https://github.com/couchbase/docs-sdk-java/pull/214/files which is now merged.

PR #212, opened 2021-03-02
DOC-8162 Java Named Scopes & Collections examples

maria-robobug:
https://issues.couchbase.com/browse/DOC-8162

- Adds an example for scoped named collections for KV, also moves the equivalent Query example code to the relevant java class.

- I also updated some parts of the examples so they are actually runnable against `travel-sample` bucket.

- There are some minor code format changes (I use VSCode, so it just formatted the java code a bit).


modules/howtos/pages/kv-operations.adoc
```diff
@@ -2,12 +2,16 @@
 :navtitle: KV Operations
 :page-topic-type: howto
 :page-aliases: document-operations.adoc,ROOT:documents-creating,ROOT:documents-updating,ROOT:documents-retrieving,ROOT:documents-deleting
+:lang: Java
+:version: 3.1.2
```
RichardSmedley: we need to use attributes.adoc to avoid this unnecessary maintenance burden.

RichardSmedley (commented):
.adoc LGTM other than comment on maintenance burden of unnecessary meta info.
Will leave the Java files +1 to @daschl

daschl (approved):

PR #213, opened 2021-03-02
Java 3.1.3

programmatix:


modules/howtos/pages/n1ql-queries-with-sdk.adoc
```diff
@@ -3,7 +3,7 @@
 :page-topic-type: howto
 :page-aliases: n1ql-query,ROOT:querying-n1ql
 :lang: Java
-:version: 3.1.2
+:version: 3.1.3
```
RichardSmedley, line 6: Would be better to pull in the attributes adoc instead of all of this monthly maintenance in multiple places.

RichardSmedley (approved):
:+1:

modules/howtos/pages/n1ql-queries-with-sdk.adoc
programmatix: I'm not even sure if these version fields are needed

RichardSmedley: Indeed, they don't seem to be atm,
but could be used to automatically update 
`The latest version (as of March 2021) is https://search.maven.org/artifact/com.couchbase.client/java-client/3.1.3/jar[3.1.3].`
and other patch version mentions
via pulling in from attributes.adoc

PR #220, opened 2021-03-23
DOC-7888 CollectionManager sample and docs

osfameron:
Run with:

  mvn -U exec:java -Dexec.mainClass="CollectionManagerExample" -Dexec.cleanupDaemonThreads=false

NB: the referred exception bug in JCBC-1792 is now marked resolved, I haven't retested with HEAD, but the example follows the expected behaviour.

### DIFF
--- a/modules/howtos/examples/CollectionManagerExample.java
+++ b/modules/howtos/examples/CollectionManagerExample.java
@@ -74,9 +74,6 @@ public class CollectionManagerExample {
         catch (ScopeExistsException e) {
           System.out.println("Scope already exists");
         }
-        catch (AuthenticationFailureException e) {
-          System.out.println("User doesn't have required privileges to administer bucket");
-        }
         // end::create-scope[]
       }
 
--- a/modules/howtos/examples/CollectionManagerExample.java
+++ b/modules/howtos/examples/CollectionManagerExample.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2021 Couchbase, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// tag::imports[]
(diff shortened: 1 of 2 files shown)


modules/howtos/examples/CollectionManagerExample.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // tag::create-scope[]
+        try {
+          collectionMgr.createScope("example-scope");
+        }
+        catch (ScopeExistsException e) {
+          System.out.println("Scope already exists");
+        }
+        catch (AuthenticationFailureException e) {
+          System.out.println("User doesn't have required privileges to administer bucket");
```
dnault: @osfameron If the user doesn't have the required permissions, the SDK throws CouchbaseException. We don't have an exception to signal an authorization failure. I'll start a discussion with the team about whether we should add one. In the mean time, should we just omit this catch block?

modules/howtos/pages/provisioning-cluster-resources.adoc
```diff
@@ -98,8 +98,51 @@ The `Flush` operation may fail if the bucket does not have flush enabled, in tha
...
+
+[source,java]
+----
+include::example$CollectionManagerExample.java[tag=drop-collection,indent=0]
+
+include::example$CollectionManagerExample.java[tag=drop-scope,indent=0]
+----
 
+Note that you require at least the https://docs.couchbase.com/server/7.0/learn/security/roles.html#bucket-admin[Bucket Admin] privilege to create or drop a Scope.
+The minimum privilege for managing Collections is https://docs.couchbase.com/server/7.0/learn/security/roles.html#manage-collections-in-scope[Manage Collections in Scope].
```
dnault: User also needs "Data Reader" (or something similar) or else they can't open the bucket to get the collections manager.

dnault (changes requested):
nice stuff!

modules/howtos/examples/CollectionManagerExample.java
osfameron: Thanks, I'll remove for now as you suggest!
(I'd seen the error while getting my login connection wrong, and made an assumption...)

modules/howtos/pages/provisioning-cluster-resources.adoc
osfameron: Thanks! In fact I'd already added data reader in the *code*, but wasn't sure that was the best practice, will document.

RichardSmedley (approved):
LGTM

osfameron:
> Run with:
> 
> mvn -U exec:java -Dexec.mainClass="CollectionManagerExample" -Dexec.cleanupDaemonThreads=false

Better command, which actually compiles it is:

    mvn compile exec:java -Dexec.mainClass=CollectionManagerExample -Dexec.cleanupDaemonThreads=false

PR #221, opened 2021-03-26
Setup test automation for Java SDK

maria-robobug:
This PR adds the test automation setup for the Java code samples, and adds 2 initial tests to ensure the setup works.
The changes are quite similar to the Go SDK [PR](https://github.com/couchbase/docs-sdk-go/pull/117).

Please note when we compile the example Java classes, I've excluded the ones in here for the moment as I am not sure why they have their own separate `pom.xml` file with version `3.0.10` of the couchbase client:
https://github.com/couchbase/docs-sdk-java/blob/release/3.1/modules/devguide/examples/java/pom.xml
Any historical reasons? I can't see many of them referenced in .adoc files, but let me know if I need to include them.

I am planning to add tests for all other remaining examples in a separate PR so that this changeset doesn't get too big.

### DIFF
--- a/modules/howtos/examples/TransactionsExample.java
+++ b/modules/howtos/examples/TransactionsExample.java
@@ -15,6 +15,13 @@
  */
 
 // tag::imports[]
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Logger;
+
 import com.couchbase.client.core.cnc.Event;
 import com.couchbase.client.core.cnc.RequestSpan;
 import com.couchbase.client.java.Bucket;
@@ -41,17 +48,11 @@ import com.couchbase.transactions.log.IllegalDocumentState;
 import com.couchbase.transactions.log.LogDefer;
 import com.couchbase.transactions.log.TransactionCleanupAttempt;
 import com.couchbase.transactions.log.TransactionCleanupEndRunEvent;
+
+import io.opentelemetry.api.trace.Span;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.time.Duration;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Optional;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.logging.Logger;
-
-import static com.couchbase.client.java.query.QueryOptions.queryOptions;
 // end::imports[]
 
 public class TransactionsExample {
@@ -78,12 +79,11 @@ public class TransactionsExample {
     static void config() {
         // tag::config[]
         Transactions transactions = Transactions.create(cluster,
-                TransactionConfigBuilder.create()
-                        .durabilityLevel(TransactionDurabilityLevel.PERSIST_TO_MAJORITY)
+                TransactionConfigBuilder.create().durabilityLevel(TransactionDurabilityLevel.PERSIST_TO_MAJORITY)
                         // tag::config_warn[]
                         .logOnFailure(true, Event.Severity.WARN)
                         // end::config_warn[]
-                    .build());
+                        .build());
         // end::config[]
     }
 
@@ -101,7 +101,7 @@ public class TransactionsExample {
                 // will be committed anyway.
                 ctx.commit();
             });
-        // tag::logging[]
+            // tag::logging[]
         } catch (TransactionCommitAmbiguous e) {
             // The application will of course want to use its own logging rather
             // than System.err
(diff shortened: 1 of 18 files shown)


modules/pom.xml
```diff
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
osfameron, line 2: original formatting was better in this case.

modules/test/scripts/init-couchbase/init-buckets.sh
```diff
@@ -0,0 +1,122 @@
...
+set -euo pipefail
+
+CB_USER="${CB_USER:-Administrator}"
+CB_PSWD="${CB_PSWD:-password}"
+
+CB_BUCKET_RAMSIZE="${CB_BUCKET_RAMSIZE:-128}"
+
+echo "couchbase-cli bucket-create travel-sample..."
+/opt/couchbase/bin/couchbase-cli bucket-create \
+-c localhost -u ${CB_USER} -p ${CB_PSWD} \
```
osfameron, line 16: Missing indentation on these continuation lines looks odd.

osfameron: (and below throughout. not sure what our coding standard is here, but I've always seen them indented.)

modules/pom.xml
maria-robobug: Not sure why, my VSCode just decided to change that, let me double check

modules/howtos/examples/TransactionsExample.java
```diff
@@ -547,19 +538,16 @@ static void deferredRollback(byte[] encoded) {
...
     static void configCleanup(byte[] encoded) {
         // tag::config-cleanup[]
-        Transactions transactions = Transactions.create(cluster, TransactionConfigBuilder.create()
-                .cleanupClientAttempts(false)
-                .cleanupLostAttempts(false)
-                .cleanupWindow(Duration.ofSeconds(120))
-                .build());
+        Transactions transactions = Transactions.create(cluster,
+                TransactionConfigBuilder.create().cleanupClientAttempts(false).cleanupLostAttempts(false)
+                        .cleanupWindow(Duration.ofSeconds(120)).build());
```
osfameron, line 550: pipeline indendation before looked cleaner

modules/test/scripts/init-couchbase/init-buckets.sh
maria-robobug: I believe it's my `bash-beautify` VSCode plugin, it formats on save (https://marketplace.visualstudio.com/items?itemName=shakram02.bash-beautify)

maria-robobug: It's been handy, but in this particular case it just seems to remove the indent, I'm not really sure what the standard is here myself. Let me switch it off and add an indent.

osfameron:
The "Ensure all classes can compile" commit has rather a lot of reformatting changes that make it hard to review, but looks sensible from a quick scan. Again I'll do a quick run at some point to check it's Hakim-proof :)

maria-robobug:
Regarding the commit, I think the problem is when I make those changes, my java extension (vscjava.vscode-java-pack) also formats the class on save. It's handy as it tells me when there are syntax issues etc... also I'm not sure if these examples have been formatted correctly previously or just differently to what my extension is doing.
https://code.visualstudio.com/docs/languages/java

I dunno if there is a way to automatically set a common style...

osfameron:
Dunno usually prettifiers are either configurable *or* the One True Standard?

If that's the standard, then no worries, but it does make it hard to review. (Doing a reformat commit, as you've done elsewhere is good.)

maria-robobug:
For next time I'll try to maybe cover formatting first(commit separately), then make changes to get the example working or whatnot. That should hopefully make it easier to look at!

modules/pom.xml
maria-robobug: Resolving as this is prettifier related like we discussed below.

modules/test/scripts/init-couchbase/init-buckets.sh
maria-robobug: I guess this is also following some kind of prettifier standard, so I might just leave it as is

PR #226, opened 2021-04-13
Add remaining bats tests for code examples

maria-robobug:
This PR adds bats tests for the remaining code examples and is a continuation of the initial merged PR [here](https://github.com/couchbase/docs-sdk-java/pull/221)

Please note that I've formatted some Java classes (VSCode plugin).
I've split the changes into relevant commits to make review a bit easier.


modules/test/scripts/init-couchbase/init-buckets.sh
```diff
@@ -13,24 +13,24 @@ CB_BUCKET_RAMSIZE="${CB_BUCKET_RAMSIZE:-128}"
...
---bucket travel-sample \
---bucket-type couchbase \
---bucket-ramsize ${CB_BUCKET_RAMSIZE} \
---bucket-replica 0 \
---bucket-priority low \
---bucket-eviction-policy fullEviction \
---enable-flush 1 \
---enable-index-replica 0 \
---wait
+    -c localhost -u ${CB_USER} -p ${CB_PSWD} \
```
osfameron, line 16: ah good, you/vscode have added that indentation where it was distressing me last time :)

maria-robobug: Yeah it was annoying me too, so I changed my plugin for formatting shell scripts: https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format

Seems much better!

modules/Makefile
```diff
@@ -31,7 +31,7 @@ tests:
 	@docker exec -t cb-test bin/bash -c "cd modules && bats -T ./test"
 
 # Run a single test.
-# i.e make TEST_NAME="Analytics" single-test
+# i.e make TEST_NAME="Analytics.java" single-test
```
osfameron, line 34: Why the new `.java` suffix in these btw?

modules/project-docs/examples/MigratingSDKCodeTo3n.java
```diff
@@ -48,19 +48,25 @@ private void init() {
     collection = bucket.defaultCollection();
   }
 
-  public void migrating_sdk_code_to_3_n_1() throws Exception { // file: project-docs/pages/migrating-sdk-code-to-3.n.adoc line: 23
+  public void migrating_sdk_code_to_3_n_1() throws Exception { // file:
+                                                               // project-docs/pages/migrating-sdk-code-to-3.n.adoc
+                                                               // line: 23
```
osfameron: this formatting is a bit odd.
Also is it just a manual denormalized comment (like we agreed we would remove from the Python examples recently) ?

modules/howtos/examples/Search.java
```diff
@@ -40,6 +40,98 @@
 import reactor.core.publisher.BaseSubscriber;
 import reactor.core.publisher.Mono;
 
+// This example assumes an index called `travel-sample-index` exists, 
+// you can create it by running the curl command below:
```
osfameron: It's good to list prereqs, but this is a very long curl command, also repeated in another java file in this commit, so at some point, this will all go out of sync.

Is it better at this point to point the reader at the init-buckets.sh script instead?

(Or, is the N1QL statement to create the index any shorter?)

osfameron:
Looks good! 👍

modules/Makefile
maria-robobug: I'd actually forgotten to put that in in the initial PR, we had that in the Go one for example, so trying to keep it consistent.

modules/project-docs/examples/MigratingSDKCodeTo3n.java
maria-robobug: Ah yeah I can probably get rid of it, VSCode kind of forced this formatting but the comments aren't really necessary indeed.

modules/howtos/examples/Search.java
maria-robobug: hmm, I think probably pointing to init-buckets.sh would be easier 👍

PR #227, opened 2021-04-16
Cert Auth

RichardSmedley:
DOC-8250

### DIFF
--- a/modules/devguide/examples/java/src/main/java/com/couchbase/devguide/CloudConnect.java
+++ b/modules/devguide/examples/java/src/main/java/com/couchbase/devguide/CloudConnect.java
@@ -16,7 +16,7 @@
 
 package com.couchbase.devguide;
 
-import com.couchbase.client.core.deps.io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+// import com.couchbase.client.core.deps.io.netty.handler.ssl.util.InsecureTrustManagerFactory; // use in testing if you use .trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)) instead of Cert Auth
 import com.couchbase.client.core.env.IoConfig;
 import com.couchbase.client.core.env.SecurityConfig;
 import com.couchbase.client.java.Bucket;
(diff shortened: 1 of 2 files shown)


daschl (changes requested):
Note that this is not certificate authentication! Cert auth is the other way round - where you provide a certificate to the server instead of user/password credentials to authenticate.

Here, we are not dealing with authentication, but the client is accepting the server certificate for the connection (you still have to provide username and password). So this is just "encrypted / tls  connections with the custom certificate the server provides)

RichardSmedley:
>  ...So this is just "encrypted / tls connections with the custom certificate the server provides)

Oops, thanks for catching that @daschl - fixed now.

modules/devguide/examples/java/src/main/java/com/couchbase/devguide/CloudConnect.java
```diff
@@ -16,7 +16,7 @@
 
 package com.couchbase.devguide;
 
-import com.couchbase.client.core.deps.io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+// import com.couchbase.client.core.deps.io.netty.handler.ssl.util.InsecureTrustManagerFactory; // use in testing if you use .trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)) instead of Cert Auth
```
daschl, line 19: still nothing to do with cert auth (comment)

modules/devguide/examples/java/src/main/java/com/couchbase/devguide/CloudConnect.java
```diff
@@ -43,7 +43,8 @@ public static void main(String... args) {
 
         ClusterEnvironment env = ClusterEnvironment.builder()
                 .securityConfig(SecurityConfig.enableTls(true)
-                        .trustManagerFactory(InsecureTrustManagerFactory.INSTANCE))
+                        // .trustManagerFactory(InsecureTrustManagerFactory.INSTANCE))  // use in testing if you use .trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)) instead of Cert Auth (below)
```
daschl, line 46: same

daschl: more accurate would be "used in testing if you want to trust all server certificates  / do not validate them"

modules/hello-world/pages/start-using-sdk.adoc
```diff
@@ -121,6 +121,9 @@ Couchbase Cloud Sample::
 +
 --
 If you are connecting to https://docs.couchbase.com/cloud/index.html[Couchbase Cloud], be sure to get the correct endpoint as well as user, password, and `couchbasecloudbucket`  -- and see the <<cloud-connections, Cloud section>>, below.
+During development, you may wish to use `.securityConfig(SecurityConfig.enableTls(true)` (commented out in the code sample below),
```
daschl, line 124: this does not make sense? enable tls is always needed both in dev and prod, I think you wanted to mention the insecure manager?

RichardSmedley:
Closing as has been done properly elsewhere.

PR #231, opened 2021-05-13
DOC-8462 - Use internal build of Couchbase Server for tests

maria-robobug:
This applies the same changes made in this PR: https://github.com/couchbase/docs-sdk-go/pull/129

I've also included a fix for the failing compilation of `AsyncOperations.java`, this was causing the `Build Code Samples` check to fail. Looks like this commit might have triggered that failure: https://github.com/couchbase/docs-sdk-java/commit/f682a63245dc5f24ebe7d392653c60730cecf769

### DIFF
--- a/modules/howtos/examples/AsyncOperations.java
+++ b/modules/howtos/examples/AsyncOperations.java
@@ -39,6 +39,7 @@ import com.couchbase.client.java.kv.GetResult;
 import io.reactivex.Single;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.util.retry.Retry;
 
 public class AsyncOperations {
 
@@ -112,7 +113,8 @@ public class AsyncOperations {
       List<String> docsToFetch = Arrays.asList("airline_1316", "airline_13391", "airline_1355");
 
       List<GetResult> results = Flux.fromIterable(docsToFetch)
-          .flatMap(key -> reactiveCollection.get(key).retryBackoff(10, Duration.ofMillis(10))).collectList().block();
+          .flatMap(key -> reactiveCollection.get(key).retryWhen(Retry.backoff(10, Duration.ofMillis(10)))).collectList()
+          .block();
       // end::retry-bulk[]
     }
 
--- a/modules/test/test-howtos.bats
+++ b/modules/test/test-howtos.bats
@@ -52,15 +52,6 @@ load 'test/test_helper.bash'
 }
 
 @test "[howtos] - KvOperations.java" {
-    # The example with tag `named-collection-upsert` fails when running against Couchbase Server 7.0.0-beta(4602) docker
-    # image with the following error:
-    #
-    # com.couchbase.client.core.error.AmbiguousTimeoutException: UpsertRequest, Reason: TIMEOUT
-
-    # Seems to be fixed when testing against Couchbase Server build `4933` locally(using vagrant), 
-    # but we have no way of running against internal builds at this point in time. 
-    skip "TODO: Example requires latest build of Couchbase Server 7.0"
-
     runExample KvOperations 
     assert_success
 }
--- a/.github/workflows/test-dev.yml
+++ b/.github/workflows/test-dev.yml
@@ -1,4 +1,4 @@
-name: Test Code Samples
+name: Test Code Samples (Dev)
 
 on:
   push:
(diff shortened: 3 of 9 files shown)


modules/test/test-howtos.bats
```diff
@@ -52,15 +52,6 @@ load 'test/test_helper.bash'
 }
 
 @test "[howtos] - KvOperations.java" {
```
maria-robobug, line 54: We no longer need to skip this as it will be tested against the latest Couchbase Server build and pass - once we merge into the default branch.

modules/test/test-howtos.bats
```diff
@@ -52,15 +52,6 @@ load 'test/test_helper.bash'
...
-    # The example with tag `named-collection-upsert` fails when running against Couchbase Server 7.0.0-beta(4602) docker
-    # image with the following error:
-    #
-    # com.couchbase.client.core.error.AmbiguousTimeoutException: UpsertRequest, Reason: TIMEOUT
-
-    # Seems to be fixed when testing against Couchbase Server build `4933` locally(using vagrant), 
-    # but we have no way of running against internal builds at this point in time. 
-    skip "TODO: Example requires latest build of Couchbase Server 7.0"
```
maria-robobug, line 62: We no longer need to skip this as it will be tested against the latest Couchbase Server build and pass - once we merge into the default branch.

maria-robobug:
Test Code Samples (GA) fails due to incompatibility with SDK v3.1.5 and the Couchbase Server version as discussed. Once merged into release 3.1 branch an additional phase called Test Code Samples (Dev) should also run (and pass), I've added status badges so that we can see this reflected on the readme.

modules/howtos/examples/AsyncOperations.java
```diff
@@ -112,7 +113,8 @@ public static void main(String... args) {
       List<String> docsToFetch = Arrays.asList("airline_1316", "airline_13391", "airline_1355");
 
       List<GetResult> results = Flux.fromIterable(docsToFetch)
-          .flatMap(key -> reactiveCollection.get(key).retryBackoff(10, Duration.ofMillis(10))).collectList().block();
+          .flatMap(key -> reactiveCollection.get(key).retryWhen(Retry.backoff(10, Duration.ofMillis(10)))).collectList()
```
maria-robobug, line 116: `retryBackoff` has been deprecated and replaced by `retryWhen`.

modules/test/test-howtos.bats
osfameron: so it will fail on initial PR, but succeed once merged, as we've talked about?

modules/howtos/examples/AsyncOperations.java
```diff
@@ -112,7 +113,8 @@ public static void main(String... args) {
...
+          .block();
```
osfameron, line 117: Is this related to the retryWhen change, or something else?
(Not important, just curious)

osfameron:
Looks good!
Re `Test Code Samples (Dev) should only run on push to default branch` commit, does this work around the behaviour we've discussed where the initial PR will fail because of lack of secrets, by only running them on the final merge?

modules/howtos/examples/AsyncOperations.java
maria-robobug: aye, `retryBackoff` was deprecated and replaced with `retryWhen`.

modules/test/test-howtos.bats
maria-robobug: Yup that's exactly it!

modules/howtos/examples/AsyncOperations.java
osfameron: sorry, I meant the `.block()` line specifically, which didn't seem *obviously* part of that API change, so just curious :)

maria-robobug:
In regards to `Test Code Samples (Dev)`, this will only run on merge/any direct push to the default branch. 
Secrets are available when running on the default branch, so the workflow can run without a hitch.

There is no point running it on fork PRs as the workflow will not have access to secrets and fail, hence why it is limited to run only when merged.

`Test Code Samples(GA)` runs tests against the public docker image of Couchbase Server, so we don't need to worry about secrets there as it is a publicly available image. This will continue to fail until we have a new public release candidate for Couchbase Server.

modules/howtos/examples/AsyncOperations.java
maria-robobug: oh, `.block()` was there before, I haven't changed that

maria-robobug: I think it's just formatting probably that might have confused things.

osfameron: ah! yes! moar coffee!

PR #237, opened 2021-06-11
DOC-7867 - SDK 3.2 Collections-first examples

maria-robobug:
[DOC-7867](https://issues.couchbase.com/browse/DOC-7867), [JCBC-1755](https://issues.couchbase.com/browse/JCBC-1755)

This PR will:

* Update all sample code to use named scopes and collections instead of defaults.
* Remove any mentions of Developer Preview
* Update Antora references of server 6.5/6.6 to 7.0
* Update references of SDK 3.1 to 3.2 where necessary.
* Add missing bats tests for `concept-docs` examples (I had totally missed these for some reason 😞)

Please note because of how big the change is, I've split the commits by doc folder so that you can look at each change separately.

The final commit is simply updating the github workflow files with the 3.2 branch so automation can run.

**Note**: The code sample tests will fail as we are using named scopes and collections for everything now, which we expect since the SDK is not compatible with the official server 7.0-beta image. I have tested them locally and they do pass, so once merged they will also run against the latest internal server build and should go green then.

### DIFF
--- a/modules/concept-docs/pages/durability-replication-failure-considerations.adoc
+++ b/modules/concept-docs/pages/durability-replication-failure-considerations.adoc
@@ -2,7 +2,7 @@
 :description: Data durability refers to the fault tolerance and persistence of data in the face of software or hardware failure.
 :page-topic-type: concept
 // :page-aliases: ROOT:failure-considerations,ROOT:durability,ROOT:enhanced-durability
-// :page-aliases: 6.6@server:sdk:durability.adoc
+// :page-aliases: 7.0@server:sdk:durability.adoc
 
 
 [abstract]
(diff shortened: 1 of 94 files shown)


modules/howtos/pages/full-text-searching-with-sdk.adoc
```diff
@@ -195,6 +196,17 @@ You can tell the Search Engine to include the full content of a certain number o
 include::example$Search.java[tag=fields,indent=0]
 ----
 
+=== Collections
```
maria-robobug, line 199: Noticed we hadn't covered this, I will create a ticket to update the other SDKs, but doing it now for Java since I am already here.

This was the change:
https://github.com/couchbase/couchbase-jvm-clients/commit/a9c325bc1cf7a3f0fb7fe08a31a4b04f1c2a8c47

modules/howtos/pages/slow-operations-logging.adoc
```diff
@@ -1,7 +1,7 @@
 = Slow Operations Logging
 :description: Tracing information on slow operations can be found in the logs as threshold logging, orphan logging, and other span metrics.
 :page-topic-type: howto
-// :page-aliases: ROOT:
+:page-aliases: ROOT:slow-operations-logging.adoc
```
RichardSmedley: not needed (no 2.7 page with this name),
but on't forget to check the two 2.7 pages are aliased in only one place each.

osfameron: I'm confused - what has this got to do with 2.7?
Perhaps we can go over the page-aliases again - all I've been doing for sanity-check is making  sure that a build with 3.1 and 3.2 runs without error.

maria-robobug: Yeah, I'd like to understand the 2.7 part a bit better and `page-aliases` in general 😓

dnault:
LGTM. There's value in keeping *some* documentation about defaultCollection(), since developers can use the 3.x SDK with pre-7.0 servers.

modules/howtos/examples/Analytics.java
```diff
@@ -86,7 +86,7 @@ public static void main(String... args) {
     {
       // tag::named[]
       AnalyticsResult result = cluster.analyticsQuery(
-        "select count(*) from airports where country = $country",
+        "select count(*) from `travel-sample`.inventory.airport where country = $country",
```
osfameron: So, this file *already* has Collection examples in it, that I did in DOC-7628.
As @dnault suggests, it might be better to keep some of the non-Collection examples.

Gonna build it now and have a look at it in context

osfameron: Yeah, as discussed on Slack, this isn't quite right - as the collections examples at the *end* are worded as if they're introducing the concept, despite it being used throughout.

Perhaps that section should be moved to top, or we should revert to Default collection examples in this case

maria-robobug: Yeah, so I think for the pages where we already have a section for named scopes and collections I will revert the examples to default as it probably makes more sense there.

modules/concept-docs/pages/durability-replication-failure-considerations.adoc
```diff
@@ -1,8 +1,8 @@
 = Durability & Failure
 :description: Data durability refers to the fault tolerance and persistence of data in the face of software or hardware failure.
 :page-topic-type: concept
-// :page-aliases: ROOT:failure-considerations,ROOT:durability,ROOT:enhanced-durability
-// :page-aliases: 7.0@server:sdk:durability.adoc
+:page-aliases: ROOT:failure-considerations,ROOT:durability,ROOT:enhanced-durability
+//:page-aliases: 7.0@server:sdk:durability.adoc
```
maria-robobug: Not sure if `7.0@server:sdk:durability.adoc` should be removed as it's a server page...? @RichardSmedley

RichardSmedley: I _think_ not only should it possibly still be there,
as it is in 3.0,
but there needs to be one the same, but with developer-guide in place of :sdk:
for content present in 4.x docs.
Not essential, though.

RichardSmedley (approved):
LGTM

PR #238, opened 2021-06-14
DOC-8615 update compatibility matrix for 7.0β

osfameron:
First pass.

### DIFF
--- a/modules/project-docs/pages/compatibility.adoc
+++ b/modules/project-docs/pages/compatibility.adoc
@@ -7,9 +7,9 @@ Plus notes on Cloud, networks, and AWS Lambda.
 [abstract]
 {description}
 
-The 3.0 SDK requires Java 8 or later to be installed, earlier versions will not work. 
-_Java 11 is recommended_, which has various enhancements like lambda local variable type inference, profiling tools, and updated security features. 
-Most of the flavors available will do, although we may only provide support for OpenJDK and Oracle JDK going forward. 
+The 3.x SDK requires Java 8 or later to be installed, earlier versions will not work.
+_Java 11 is recommended_, which has various enhancements like lambda local variable type inference, profiling tools, and updated security features.
+Most of the flavors available will do, although we may only provide support for OpenJDK and Oracle JDK going forward.
 The Java SDK _has not been tested (and is not supported on) Java 12 or later_.
 
 
@@ -26,9 +26,9 @@ It is best to upgrade either the SDK or the Couchbase version you are using.
 
 .Recommended SDK per Server Version Matrix
 [#table_sdk_versions]
-[cols="40,25,20,20,25"]
+[cols="40,25,20,20,30"]
 |===
-| | SDK 2.4, 2.5 | SDK 2.6 | SDK 2.7  | SDK 3.0, 3.1 
+| | SDK 2.4, 2.5 | SDK 2.6 | SDK 2.7  | SDK 3.0, 3.1
 
 | *Server 5.0*
 | *✔*
@@ -58,8 +58,17 @@ It is best to upgrade either the SDK or the Couchbase version you are using.
 | *✖*
 | *✖*
 | *◎*
-| *✔*
+| *◎* _(see below)_
 |===
+
+[IMPORTANT]
+====
+An API change in the planned release version of Couchbase Server 7.0 means changes in the latest versions of the SDK which render these new SDK versions incompatible with the beta versions of 7.0.
+
+Java SDK *3.1.3* is the latest version compatible with Server 7.0β.
+
+Later SDK 3.1.x releases will be limited to the Default Scope and Collection with the beta 7.0 server, and will only be fully compatible with the full release version.
+====
 
 Note the https://www.couchbase.com/support-policy[End of Life dates^] for Couchbase Server and SDK versions.
 See the notes there for Support details.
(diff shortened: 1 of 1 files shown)


osfameron:
(apologies, a few whitespace artefacts, view diff with `-w` / whitespace omitted.

modules/project-docs/pages/compatibility.adoc
```diff
@@ -58,9 +58,12 @@ It is best to upgrade either the SDK or the Couchbase version you are using.
 | *✖*
 | *✖*
 | *◎*
-| *✔*
+| *◎* _(see below)_
 |===
 
+IMPORTANT: when running against Server 7.0β, Java SDK *3.1.6* is the latest compatible version.
+Later SDK 3.1.x releases will only be able to use the Default Collection, due to a change of API.
+
```
RichardSmedley, line 72: I think we should outline the change and future plan first (for reasons discussed), then discuss the versions for beta users. 
How about:
"IMPORTANT: An API change in the planned release version of Couchbase Server 7.0 means changes in the latest versions of the SDK which render these new SDK versions incompatible with the _beta_ versions of 7.0.
Java SDK *3.1.6* is the latest version compatible with Server 7.0β.
Later SDK 3.1.x releases will only be fully compatible with the full release version of 7.0, except in the case of just running against _default_ Scopes and Collections with the beta 7.0 Server."
Could make this a partial (once settled on) and pull it in to each compat doc.

modules/project-docs/pages/compatibility.adoc
```diff
@@ -28,7 +28,7 @@ It is best to upgrade either the SDK or the Couchbase version you are using.
 [#table_sdk_versions]
 [cols="40,25,20,20,25"]
```
RichardSmedley: How about [cols="30,25,20,20,35"] -to make room for the 
*◎* _(see below)_

modules/project-docs/pages/compatibility.adoc
```diff
@@ -7,9 +7,9 @@ Plus notes on Cloud, networks, and AWS Lambda.
 [abstract]
 {description}
 
-The 3.0 SDK requires Java 8 or later to be installed, earlier versions will not work. 
-_Java 11 is recommended_, which has various enhancements like lambda local variable type inference, profiling tools, and updated security features. 
-Most of the flavors available will do, although we may only provide support for OpenJDK and Oracle JDK going forward. 
+The 3.0 SDK requires Java 8 or later to be installed, earlier versions will not work.
```
RichardSmedley: Worth changing to 3.x here.

modules/project-docs/pages/compatibility.adoc
osfameron: yeah, you mentioned expanding the column - I noticed that because the header "SDK 3.0, 3.1" is already fairly long, it "looked fine on my device"...

(e.g. I'm not sure how best to test what the optimal width is? Happy to go with 35 and your intuition of course.)

modules/project-docs/pages/compatibility.adoc
osfameron: Sounds good! Bringing in @AV25242 to give his opinion on this and rest of PR.

modules/project-docs/pages/compatibility.adoc
RichardSmedley: That was intuition, I didn't have time to test - but one of us will have to before it reaches prod ;)

osfameron: OK, in that case, let's say my local test is Good Enough For Staging, and we'll see there from a few different browsers if it needs tweaking?

modules/project-docs/pages/compatibility.adoc
```diff
@@ -58,9 +58,12 @@ It is best to upgrade either the SDK or the Couchbase version you are using.
 | *✖*
 | *✖*
 | *◎*
-| *✔*
+| *◎* _(see below)_
 |===
 
+IMPORTANT: when running against Server 7.0β, Java SDK *3.1.6* is the latest compatible version.
```
AV25242: @osfameron  3.1.6 is the latest compatible is not right.

refer to this page https://www.couchbase.com/products/developer-sdk/compatibility
for instance Java is compatible with SDK versions ranging from SDK 3.0.9 through 3.1.3
Likewise add this comment as to why using the range of versions 
 : **The latest release of Couchbase SDKs include changes for the upcoming GA release. As a result before GA you will need to use a specific range of releases.**

osfameron: @AV25242 thanks - I'm confused though - the internal google doc linked to the ticket does mention 3.1.6, should I ignore that document in favour of the page you link above, in all cases?

modules/project-docs/pages/compatibility.adoc
AV25242: > Sounds good! Bringing in @AV25242 to give his opinion on this and rest of PR.

I had mentioned it above but 3.1.6 for instance is not the one thats compatible with 7.0 Beta refer to the currently compatibility matrix page 
   https://www.couchbase.com/products/developer-sdk/compatibility
As an example for Java SDK versions ranging from SDK 3.0.9 through 3.1.3 will be compatible

osfameron: Thanks - is that document correct for all cases?

We were looking at the developer-updated GDoc and taking the complement (e.g. assuming that what works for Beta does NOT work for GA and v.v.)

Does this mean that 3.1.4-3.1.5 don't work correctly with either 7.0beta *or* 7.0?

osfameron:
Force pushed a new version - @RichardSmedley @AV25242 how's this?

osfameron:
@AV25242 has given his lgtm. Merging.

PR #239, opened 2021-06-23
Remove outstanding Developer Preview references

maria-robobug:
Removes any outstanding mentions of `Developer Preview` and `[.status]#Developer Preview#` tags.

### DIFF
--- a/modules/howtos/pages/observability-metrics.adoc
+++ b/modules/howtos/pages/observability-metrics.adoc
@@ -1,6 +1,5 @@
 = Observability Metrics Reporting
 :description: Individual request tracing presents a very specific (though isolated) view of the system.
-:page-status: Developer Preview
 :page-topic-type: howto
 
 [abstract]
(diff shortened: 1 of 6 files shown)


modules/howtos/pages/observability-metrics.adoc
```diff
@@ -22,15 +21,8 @@ Right now only the first category is implemented by the SDK, more are planned.
 
 
 == The Default AggregatingMeter
-[.status]#Developer Preview#
 
 The default implementation aggregates and logs request and response metrics.
-Since the feature is in developer preview, you need to manually enable it on the environment:
```
maria-robobug, line 28: I removed this because it mentions _developer preview_ and tells you to enable it through the SDK.

I guess this would be enabled by default if the user has Couchbase Server 7.0? So I figured you won't need to show an example of enabling this anymore. But let me know if we should keep this.

RichardSmedley: I think you're right.

RichardSmedley (approved):
LGTM

PR #247, opened 2021-07-19
A sweep on transactions docs for 7.0 release, particularly addressing…

programmatix:
… query and custom metadata collections.

Must be staged together with a related PR to docs-sdk-common.

### DIFF
--- a/modules/howtos/examples/TransactionsExample.java
+++ b/modules/howtos/examples/TransactionsExample.java
@@ -29,18 +29,24 @@ import com.couchbase.client.java.Cluster;
 import com.couchbase.client.java.Collection;
 import com.couchbase.client.java.Scope;
 import com.couchbase.client.java.ReactiveCollection;
+import com.couchbase.client.java.ReactiveScope;
+import com.couchbase.client.java.Scope;
+import com.couchbase.client.java.json.JsonArray;
 import com.couchbase.client.java.json.JsonObject;
 import com.couchbase.client.java.kv.GetResult;
 import com.couchbase.client.java.query.QueryOptions;
 import com.couchbase.client.java.query.QueryProfile;
 import com.couchbase.client.java.query.QueryResult;
+import com.couchbase.client.java.query.QueryScanConsistency;
 import com.couchbase.client.tracing.opentelemetry.OpenTelemetryRequestSpan;
+import com.couchbase.transactions.SingleQueryTransactionResult;
 import com.couchbase.transactions.TransactionDurabilityLevel;
 import com.couchbase.transactions.TransactionGetResult;
 import com.couchbase.transactions.TransactionQueryOptions;
 import com.couchbase.transactions.TransactionResult;
 import com.couchbase.transactions.Transactions;
 import com.couchbase.transactions.config.PerTransactionConfigBuilder;
+import com.couchbase.transactions.config.SingleQueryTransactionConfigBuilder;
 import com.couchbase.transactions.config.TransactionConfigBuilder;
 import com.couchbase.transactions.deferred.TransactionSerializedContext;
 import com.couchbase.transactions.error.TransactionCommitAmbiguous;
@@ -64,7 +70,7 @@ public class TransactionsExample {
     public static void main(String... args) {
         // tag::init[]
         // Initialize the Couchbase cluster
-        Cluster cluster = Cluster.connect("localhost", "Administrator", "password");
+        Cluster cluster = Cluster.connect("localhost", "username", "password");
         Bucket bucket = cluster.bucket("travel-sample");
         Scope scope = bucket.scope("inventory");
         Collection collection = scope.collection("airport");
@@ -76,6 +82,8 @@ public class TransactionsExample {
         TransactionsExample.cluster = cluster;
         TransactionsExample.collection = collection;
         TransactionsExample.transactions = transactions;
+
+        queryExamples();
     }
 
     static void config() {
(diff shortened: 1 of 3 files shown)


modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -105,17 +102,12 @@ include::example$TransactionsExample.java[tag=createReactive,indent=0]
 ----
 
 The synchronous mode is the easiest to write and understand.
-The asynchronous API allows you to build your application in a reactive style, without large thread pools, which can help you scale with excellent efficiency.
+The asynchronous API allows you to build your application in a reactive style, which can help you scale with excellent efficiency.
 Those new to reactive programming may want to check out https://projectreactor.io/[the Project Reactor site] for more details on this powerful paradigm.
 
-NOTE: Some `AttemptContextReactive` methods, notably `remove`, return `Mono<Void>`.
```
programmatix, line 111: Moved this bit into the Remove section where it's most relevant.

RichardSmedley (approved):
:+1:

modules/howtos/examples/TransactionsExample.java
```diff
@@ -64,7 +70,7 @@
     public static void main(String... args) {
         // tag::init[]
         // Initialize the Couchbase cluster
-        Cluster cluster = Cluster.connect("localhost", "Administrator", "password");
+        Cluster cluster = Cluster.connect("localhost", "username", "password");
```
ingenthr, line 73: thanks for fixing this anti-pattern.  :)

modules/howtos/examples/TransactionsExample.java
```diff
@@ -184,6 +193,18 @@ static void examples() {
...
                 ctx.remove(docC);
 
+                // Performing a SELECT N1QL query against a scope:
+                QueryResult qr = ctx.query(inventory, "SELECT * FROM hotel WHERE country = $1",
+                        TransactionQueryOptions.queryOptions()
+                                .parameters(JsonArray.from("United Kingdom")));
+                List<JsonObject> rows = qr.rowsAs(JsonObject.class);
+
+                // Performing an UPDATE N1QL query against a scope:
+                ctx.query(inventory, "UPDATE hotel SET price = $1 WHERE name = $2",
```
ingenthr, line 203: nitpick, but I would think it might be better to have something that updates multiple documents since that's a bit more realistic to where someone would likely use an UPDATE WHERE.  in fact, I might even mention this in a comment.

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -393,15 +449,15 @@ include::example$TransactionsExample.java[tag=full-logging,indent=0]
...
+[.status]#Developer Preview#
 
-The 1.1.6 release introduces support for OpenTelemetry and OpenTracing.
-
-If the underlying Couchbase Java SDK is configured for OpenTelemetry or OpenTracing, then transaction spans will now be output automatically.
+The 1.1.6 release introduces support, at a Developer Preview level, for detailed tracing.
+This telemetry is particularly useful for monitoring performance.
 
-See the xref:howtos:observability-tracing.adoc[Couchbase Java SDK Request Tracing documentation] for more guidance on configuring the Couchbase Java SDK appropriately.
+Note: Developer Preview here means the feature works and is safe to use, but the naming of spans and attributes, and which spans are created, may change before this feature is generally available.
```
ingenthr, line 457: you could refer to the interface taxonomy as uncommitted.

modules/howtos/examples/TransactionsExample.java
```diff
@@ -217,9 +240,24 @@ static void examplesReactive() {
...
+                    .then(ctx.query(inventory, "SELECT * FROM hotel WHERE country = $1",
+                            TransactionQueryOptions.queryOptions()
+                                    .parameters(JsonArray.from("United Kingdom"))))
+
+                    .doOnNext(queryResult -> queryResult.rowsAs(JsonObject.class)
+                            .doOnNext(row -> {
+                                // the application would do something with each row here
+                            }))
+
+                    // Performing an UPDATE N1QL query against a scope:
```
ingenthr, line 256: might add to this comment "against a scope containing a collection named `hotel`" or something like this

ingenthr (commented):
I think my primary criticism would be that the examples might be too simple.  For example, might there be a way to update multiple documents?  and maybe have any reads involved in the update?

For example, imagine doing a select to find hotels with a particular attribute, like part of a hotel chain, then asking a pure java function for the price, then updating the hotel chain.  The pure function could be something like consultMlModel() where internally it's just random.

What you have here is great.  I think it can be improved a bit by keeping it simple, but also showing something powerful.

RichardSmedley:
Thanks @ingenthr 
Will look at these (and the other repo's comments) with Graham,
and try and get some changes in in next few days  :+1:

modules/howtos/examples/TransactionsExample.java
programmatix: I thought you'd like this :)

modules/howtos/examples/TransactionsExample.java
programmatix: Yes, fair point

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
programmatix: Ah well it's deliberately not as the only API is parentSpan(RequestSpan span) in PerTransactionConfigBuilder and this isn't going to change - RequestSpan is nailed down.

modules/howtos/examples/TransactionsExample.java
programmatix: Will do

programmatix:
Thanks for the feedback Matt.  I was thinking the same (N1QL examples need to be more realistic), I'd just spent a bunch of time on this and wanted to get the improvements up, before taking another swing at it.  Those are good suggestions, I'll fold something like it in.

PR #249, opened 2021-07-20
3.2.0 release

programmatix:


programmatix:
I think the build failure is because I was parallelising releasing with release notes.  If there's a way to retrigger it, it will likely succeed.

RichardSmedley (approved):
:+1:

PR #256, opened 2021-08-05
Automation improvements for Github Actions

maria-robobug:
This PR adds the following automation updates:

- Send slack notification to #docs-automation when a workflow fails on the **default** branch (i.e release/3.2)
- Update couchbase server version to 7.0.0 GA for testing
- Mark `build` and `test` workflows as successful if the PR changes don't actually require any automation to run (i.e changing an .adoc file).  Previously the workflows just wouldn't run, which makes sense, however this doesn't work with Github branch protection as there is no option for conditional status checks. Github will always expect those checks to run.
To mitigate this, we skip the steps in the workflow and mark it as successful, allowing us to merge the PR with no issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,20 +1,11 @@
-name: Build Code Samples 
+name: Build Code Samples
 
 on:
   push:
-    branches: [release/3.2]
-    paths:
-      - 'modules/test/*'
-      - 'modules/Makefile'
-      - '**/*.java'
-      - '**/pom.xml'
+    branches: [ release/3.* ]
   pull_request:
-    branches: [release/3.2]
-    paths:
-      - 'modules/test/*'
-      - 'modules/Makefile'
-      - '**/*.java'
-      - '**/pom.xml'
+    branches: [ release/3.* ]
+
 env:
   JAVA_VERSION: 15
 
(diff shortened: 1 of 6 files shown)


maria-robobug:
Not fixing the failures in this PR because I want to see the notification sent to the slack channel when this is merged.
Will probably do a separate PR to fix the actual failure.

.github/workflows/build.yml
```diff
@@ -1,20 +1,10 @@
...
 on:
   push:
-    branches: [release/3.2]
-    paths:
-      - 'modules/test/*'
-      - 'modules/Makefile'
-      - '**/*.java'
-      - '**/pom.xml'
+    branches: [ release/3.* ]
   pull_request:
```
osfameron, line 6: is this line needed? seems not to have any child keys

osfameron: (same in at least one other .yml file, I won't repeat the comment this time 😄 )

osfameron (approved):
Looks good to me!
(I don't know Github workflows, so take with appropriate pinch of salt, but looks sensible and well commented where required, which is really helpful.)

Was going to ask about build status, but I see you've mentioned that in comment, so 👍

.github/workflows/build.yml
maria-robobug: I was trying to see if not specifying the branch would still work, which it does, I don't think it will allow me to do `pull_request` without the `:` at the end.

Alternatively, I could just add `branches: [ release/3.* ]` in there too to be explicit (like before). Probably what I *should* do, I'll update. Was trying to be too clever but it just looks odd.

PR #260, opened 2021-08-12
JCBC-1855: Update Cloud Connect sample code

daschl:
Also some other small fixes to make sure the code runs in an
IDE.

Note that the cloud code does not use the travel sample data, since it is not available in cloud. So it starts from an empty bucket and  should work out of the box.


maria-robobug (approved):
👍 LGTM, thanks for fixing the failing examples! One less thing for me to look at 🥳

PR #261, opened 2021-08-12
Ensure workflows are skipped when path filters don't apply

maria-robobug:
To satisfy branch protection rules on Github, we need to ensure that
the `build` and `test` status checks can be skipped if code related files are not
changed. The initial github Action I used for this was marking the
status checks as passed, which isn't entirely correct and gives users false
positives.

Using https://github.com/marketplace/actions/skip-duplicate-actions
we can run an initial job that will check the provided path filters
to determine if subsequent jobs should be "skipped" if the filters don't apply.
This action also gives us the ability to cancel outdated workflow runs
when a new commit is pushed.

I also added some minor improvements on how we specify what branch the workflows
should run: https://github.com/couchbase/docs-sdk-java/compare/release/3.2...maria-robobug:DOC-8932?expand=1#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R5.
This should avoid the need to change each workflow yml for every minor release,
note that I start from 3.1 because that's when we started our automation journey! 😄


.github/workflows/build.yml
```diff
@@ -2,15 +2,35 @@ name: Build Code Samples
 
 on:
   push:
-    branches: [ release/3.* ]
+    branches: [ "release/3.[1-9]" ]
   pull_request:
-    branches: [ release/3.* ]
+    branches: [ "release/3.[1-9]" ]
```
osfameron: Does 3.0 cause us actual problems?
Just in case we get up to 3.10 it's probably best to add * at the end.

maria-robobug: I think not actually, because as long as this is not actually in the 3.0 code branch then it shouldn't matter, let me change that.

osfameron: Good call, if it isn't needed, then pattern-matching on numbers can be a bit icky 🤢

osfameron (approved):
Minor comment, but looks good.
(Though I don't understand the github workflows language particularly, so if you want to go through it in more depth on Zoom, that'd be interesting, maybe when Richard is back? But all looks superficially sensible 😄 )

maria-robobug:
No worries, can go over it on Zoom when Richard is back yeah!

FYI I have to push a commit to test the "skipping" status checks part also, so you may see a random `Test skipping status check` in a little bit.

maria-robobug:
Alright, it works, will revert my tests! 🥳

PR #263, opened 2021-08-13
AggregatingMeter has been renamed to LoggingMeter in 3.2

maria-robobug:
The example was updated in this [PR](https://github.com/couchbase/docs-sdk-java/pull/260/files#diff-c66bad1c2211d722d885f9f80edf4a83c482cd6623b94b4550c72af790cf17f7R20 ) already, so just updating the adoc.

### DIFF
--- a/modules/howtos/pages/observability-metrics.adoc
+++ b/modules/howtos/pages/observability-metrics.adoc
@@ -20,7 +20,7 @@ Metrics broadly fall into the following categories:
 Right now only the first category is implemented by the SDK, more are planned.
 
 
-== The Default AggregatingMeter
+== The Default LoggingMeter
 
 The default implementation aggregates and logs request and response metrics.
 
@@ -31,7 +31,7 @@ By default the metrics will be emitted every 10 minutes, but you can customize t
 include::example$Metrics.java[tag=metrics-enable-custom,indent=0]
 ----
 
-Once enabled, there is no further configuration needed. The `AggregatingMeter` will emit the collected request statistics every interval. 
+Once enabled, there is no further configuration needed. The `LoggingMeter` will emit the collected request statistics every interval.
 A possible report looks like this (prettified for better readability):
 
 [source,json]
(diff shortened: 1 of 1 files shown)


osfameron (approved):
lgtm

modules/howtos/pages/observability-metrics.adoc
```diff
@@ -20,7 +20,7 @@ Metrics broadly fall into the following categories:
 Right now only the first category is implemented by the SDK, more are planned.
 
 
-== The Default AggregatingMeter
+== The Default LoggingMeter
 
 The default implementation aggregates and logs request and response metrics.
```
RichardSmedley, line 29: Should we mention that it has a different name since 3.1?

maria-robobug: Good shout, I'll add a note

RichardSmedley: Thanks  :+1:

modules/howtos/pages/observability-metrics.adoc
```diff
@@ -19,6 +19,10 @@ Metrics broadly fall into the following categories:
 
 Right now only the first category is implemented by the SDK, more are planned.
 
+[NOTE]
+====
+The `AggregatingMeter`, which was previously available in SDK 3.1, has been renamed to `LoggingMeter` in the 3.2 release. The class is marked with `@Stability.Volatile` and may be subject to further change in later releases.
```
maria-robobug: Just double checking this with @daschl, I'm unsure if this should still be considered "Developer Preview" since the class is marked as Volatile.

maria-robobug: Looks like `LoggingMeter` shouldn't be marked as volatile, so I will remove any mention of this. 
Bug raised on the SDK side: https://issues.couchbase.com/browse/JVMCBC-1007

PR #284, opened 2021-10-18
Fix document-operations.adoc alias

maria-robobug:
https://issues.couchbase.com/browse/DOC-9301

### DIFF
--- a/modules/howtos/pages/kv-operations.adoc
+++ b/modules/howtos/pages/kv-operations.adoc
@@ -2,7 +2,7 @@
 :description: Key Value (KV) or data service offers the simplest way to retrieve or mutate data where the key is known.
 :navtitle: KV Operations
 :page-topic-type: howto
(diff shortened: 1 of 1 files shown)


osfameron (commented):
Looks OK, I'd like to see `DOC-9301` in the commit subject though.

As I understand it, this means that https://docs.couchbase.com/java-sdk/current/document-operations.html will resolve (currently 404, while the 2.7 doc works)

maria-robobug:
Hmm, no, it should go to the  KV operations page as we don't have `document-operations.adoc` in 3.x docs.
https://docs.couchbase.com/java-sdk/current/howtos/kv-operations.html

osfameron:
OK, but that *url* will resolve to the page you've modified.
All good, just I find this kind of `COME FROM` thing confusing, so like to double check 🤣

PR #289, opened 2021-11-02
DOC-8913: Configuration Settings rewording

osfameron:
The "tables" referred to are really sections.
Slightly updated wording, normalised the sections to h3,
linked to each configuration section from the intro text.

### DIFF
--- a/modules/ref/pages/client-settings.adoc
+++ b/modules/ref/pages/client-settings.adoc
@@ -3,6 +3,7 @@
 :nav-title: Client Settings
 :page-topic-type: reference
 :page-aliases: ROOT:client-settings, ROOT:env-config
+:page-toclevels: 2
 
 [abstract]
 {description}
--- a/modules/ref/pages/client-settings.adoc
+++ b/modules/ref/pages/client-settings.adoc
@@ -1,5 +1,5 @@
 = Client Settings for the Java SDK
-:description: pass:q[The `ClusterEnvironment` class enables you to configure Java SDK options for bootstrapping, timeouts, reliability, and performance.]
+:description: pass:q[The `ClusterEnvironment` class enables you to configure Java SDK options for security, timeouts, reliability, and performance.]
 :nav-title: Client Settings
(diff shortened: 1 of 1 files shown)


modules/ref/pages/client-settings.adoc
```diff
@@ -140,7 +146,7 @@ As an alternative to specifying the certificates to trust, you can specify a cus
 See the xref:howtos:managing-connections.adoc#ssl[Connection Management] section for more details on how to set it up properly.
 
 
-== I/O Options
+=== I/O Options
```
RichardSmedley, line 150: You're losing most of the options from the rh-nav?

osfameron: Ah! Nope, hadn't noticed that...

But some options were h2, some h3, and didn't seem to be a hierarchy, so made sense to normalise them.

Could remove the "== Configuration Options" section entirely, and promote the children to h2?

RichardSmedley: h3 will be fine once we have it as a discretionary option for rh-nav.
Meanwhile h2 including the config options was just something of a least-worst compromise.

An alternative would be to use [discrete] to remove config options from the notional hierarchy but I don't think that's any better.

there is definitely a grouping there, so removing that section feels a little odd....
although I could be persuaded (as we will end up with a compromise for now, whatever is done).

RichardSmedley: Okay, let's give it a go.

RichardSmedley (approved):
Trying with h3s  :+1:

osfameron:
Fixed TOC menu:

Screenshot 2021-11-03 at 17 28 22

This requires:
* https://github.com/couchbase/docs-sdk-common/pull/36
* https://github.com/couchbase/docs-ui/pull/126

osfameron:
Merging as docs-ui and docs-site (and docs-sdk-common) pieces are all ready to go

PR #290, opened 2021-11-04
Rayoffiah adding student tutorial code

RayOffiah:


modules/student/examples/pom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
maria-robobug, line 1: Hmm, we already have a pom file for all the examples, is there a reason we need another one?

modules/test/scripts/init-couchbase/init-buckets.sh
```diff
@@ -76,4 +76,61 @@ until curl --fail -s -u ${CB_USER}:${CB_PSWD} http://${CB_HOST}:8094/api/index/t
...
+    --enable-index-replica 0 \
+    --wait
+
+sleep 5
+
+echo "creating art-school-scope ..."
+curl -X POST --fail -s -u ${CB_USER}:${CB_PSWD} \
+    http://${CB_HOST}:8091/pools/default/buckets/student-bucket/scopes \
+    -H "Content-Type: application/x-www-form-urlencoded" \
+    -d name=art-school-scope \
```
maria-robobug: Don't think we need the ending `\` here as that's the last line of the command.

modules/test/scripts/init-couchbase/init-buckets.sh
```diff
@@ -76,4 +76,61 @@ until curl --fail -s -u ${CB_USER}:${CB_PSWD} http://${CB_HOST}:8094/api/index/t
...
+
+sleep 5
+
+echo "creating student record collection ..."
+curl -X POST --fail -s -u ${CB_USER}:${CB_PSWD} \
+    http://${CB_HOST}:8091/pools/default/buckets/student-bucket/scopes/art-school-scope/collections \
+    -H "Content-Type: application/x-www-form-urlencoded" \
+    -d name=student-record-collection \
```
maria-robobug: Same here, remove `\`.

modules/test/scripts/init-couchbase/init-buckets.sh
```diff
@@ -76,4 +76,61 @@ until curl --fail -s -u ${CB_USER}:${CB_PSWD} http://${CB_HOST}:8094/api/index/t
...
+
+sleep 5
+
+echo "creating course record collection ..."
+curl -X POST --fail -s -u ${CB_USER}:${CB_PSWD} \
+    http://${CB_HOST}:8091/pools/default/buckets/student-bucket/scopes/art-school-scope/collections \
+    -H "Content-Type: application/x-www-form-urlencoded" \
+    -d name=course-record-collection \
```
maria-robobug: Same here, remove `\`.

modules/test/test-student.bats
```diff
@@ -0,0 +1,38 @@
+#!./test/libs/bats/bin/bats
+
+load 'test/test_helper.bash'
+
+@test "[student] - WipeCollections.java" {
```
maria-robobug: It would be more idiomatic to put this in a `setup_file()` function I think, this is a bats hook that will be guaranteed to run before any of the tests in this file.

https://bats-core.readthedocs.io/en/stable/writing-tests.html?highlight=hooks#setup-and-teardown-pre-and-post-test-hooks

So maybe something like this instead of a `@test`:
```
setup_file() {
    runExample WipeCollections
    [ "$status" -eq 0 ] \\ (I'm not sure the setup_file() function has access to `assert_success` so I used this directly instead)
}
```

Alternatively, you could just remove the class and test as we don't need to worry about clearing the db when CI runs. We always have a fresh new container when the tests run. 

See how the first approach works. If not there's the fallback of not worrying about it 😄 .

modules/student/examples/InsertCourses.java
```diff
@@ -0,0 +1,39 @@
...
+        JsonObject course = JsonObject.create()
+                .put("course-name", name)
+                .put("faculty", faculty)
+                .put("credit-points", creditPoints)
+                .put("type", "course");
+
+        collection.upsert(id, course);
+
+    }
+}
```
maria-robobug: Missing newline here.

modules/test/test-student.bats
```diff
@@ -0,0 +1,38 @@
...
+
+@test "[student] - ArtSchoolRetrieverParameters.java" {
+    runExample ArtSchoolRetrieverParameters
+    assert_success
+}
+
+@test "[student] - AddEnrollments.java" {
+    runExample AddEnrollments
+    assert_success
+}
```
maria-robobug: Missing newline.

modules/student/examples/AddEnrollments.java
```diff
@@ -0,0 +1,77 @@
...
+        JsonObject graphic_design = retrieveCourse(cluster, "graphic design");    // <5>
+        JsonObject art_history = retrieveCourse(cluster, "art history");    // <5>
+
+        String currentDate = LocalDate.now().format(DateTimeFormatter.ISO_DATE);     // <6>
+
+        // Create Hilary's enrollments
+
+        JsonArray enrollments = JsonArray.create();     // <7>
+
+        enrollments.add(JsonObject.create().put("course-id", graphic_design.getString("id"))
```
maria-robobug: Could we have the `.put` on the next line for better readability?

modules/student/examples/AddEnrollments.java
```diff
@@ -0,0 +1,77 @@
...
+                .put("date-enrolled", currentDate));    // <8>
+
+        enrollments.add(JsonObject.create().put("course-id", art_history.getString("id"))
```
maria-robobug: Same for this `.put`

maria-robobug (changes requested):
Looking good! Just wanted a few minor changes.

The tests are failing because of `Search.java` which has tended to be a bit flaky, I need to fix that. Usually a re-run of the workflow should fix it, but I will try and sort this soon as it happens quite frequently now.

modules/student/examples/pom.xml
RayOffiah: Yes, the tutorial uses a pom file to get the dependencies so folk can get up and running quickly. I put it here just to make sure I have a copy of it outside of the docs.

maria-robobug: Ok, just thinking whenever we have a new SDK release this pom will need to be updated as well to make sure we always keep the Couchbase SDK version up to date.

For example, the latest SDK version is now 3.2.2...so that will need to be updated in this pom also.
https://github.com/couchbase/docs-sdk-java/blob/release/3.2/modules/pom.xml#L10-L14

modules/test/test-student.bats
RayOffiah: Handy to know :-)

modules/student/examples/pom.xml
RayOffiah: Done!

Yup, not ideal, but I'm keen that they just have the smallest number of files to deal with to get up and running.

osfameron: Can we generate the pom files in antora... 🤔

maria-robobug (approved):

PR #292, opened 2021-11-11
Removing 'type' field following review. Not a recommended practice.

RayOffiah:
Removing the 'type' field from the example code. Now we have collections there's no need to attach a type to a document.


maria-robobug (approved):
👍 LGTM

PR #295, opened 2021-12-08
Add kv examples for dev-reorg

maria-robobug:
Supplements https://github.com/couchbase/docs-server/pull/2275 

Adding some basic examples that we can use in the `dev-reorg` server docs: https://issues.couchbase.com/browse/DOC-9205

### DIFF
--- a/modules/hello-world/examples/KvBulkHelloWorld.java
+++ b/modules/hello-world/examples/KvBulkHelloWorld.java
@@ -12,7 +12,7 @@
  * the License.
  */
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import com.couchbase.client.java.Bucket;
 import com.couchbase.client.java.Cluster;
@@ -36,10 +36,11 @@ public class KvBulkHelloWorld {
     JsonObject user2 = JsonObject.create().put("id", "user_222").put("email", "jerry_mouse@gmail.com");
     JsonObject user3 = JsonObject.create().put("id", "user_333").put("email", "mickey_mouse@gmail.com");
 
-    List<JsonDocument> documents = new ArrayList<>();
-    documents.add(new JsonDocument("user_111", user1));
-    documents.add(new JsonDocument("user_222", user2));
-    documents.add(new JsonDocument("user_333", user3));
+    List<JsonDocument> documents = Arrays.asList(
+      new JsonDocument("user_111", user1),
+      new JsonDocument("user_222", user2),
+      new JsonDocument("user_333", user3)
+    );
     // end::kv-users[]
 
     {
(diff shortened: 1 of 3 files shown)


osfameron (commented):
does merging release/3.2 into the PR apply cleanly when we merge the PR?

(I always rebase, as it feels less messy, but obviously doing it in github interface is a win too if it's not too confusing...)

modules/hello-world/examples/KvBulkHelloWorld.java
```diff
@@ -0,0 +1,152 @@
...
+        System.out.println("CAS: " + result.cas());
+      }
+      // end::kv-bulk-remove[]
+    }
+  }
+}
+
+
+// tag::kv-bulk-class[]
+class JsonDocument {
```
osfameron, line 132: Is it best practice to define a class to store the tuple of id/value?

(e.g. as opposed to just using a Map and iterating over key/values?
Hmm, it's Java, so I guess an explicit new class may well be idiomatic... just seemed a bit overkill)

modules/hello-world/examples/KvHelloWorldScoped.java
```diff
@@ -0,0 +1,224 @@
...
+    Bucket bucket = cluster.bucket("travel-sample");
+    Collection hotelCollection = bucket.scope("inventory").collection("hotel");
+
+    {
+      System.out.println("Example: [kv-insert]");
+      // tag::kv-insert[]
+      // Create the document object.
+      JsonObject geo = JsonObject.create()
+          .put("lat", 51.35785)
+          .put("lon", 0.55818).put("accuracy", "RANGE_INTERPOLATED");
```
osfameron: final `.put` on a new line for consistency

modules/hello-world/examples/KvHelloWorldScoped.java
```diff
@@ -0,0 +1,224 @@
...
+      // Print the result's CAS metadata to the console.
+      System.out.println("CAS:" + upsertResult.cas());
+      // end::kv-update-upsert[]
+    }
+
+    {
+      System.out.println("Example: [kv-update-subdoc]");
+      // tag::kv-update-subdoc[]
+      List<MutateInSpec> specs = new ArrayList<>();
+      specs.add(MutateInSpec.upsert("pets_ok", true));
```
osfameron: I've seen (and used) `= Arrays.asList(MutateInSpec.upsert(...))` elsewhere instead.

modules/test/test-hello-world.bats
```diff
@@ -19,3 +19,13 @@ load 'test/test_helper.bash'
     assert_success
     assert_output --partial "RESULT: bar"
 }
+
+@test "[hello-world] - KvHelloWorldScoped.java" {
```
osfameron, line 23: ah, I reordered some test examples in Howtos yesterday, as I was expecting them to be in Alphabetical order.
I see here that's not the case 😄

osfameron (approved):
Looks good 👍

modules/hello-world/examples/KvBulkHelloWorld.java
osfameron: 💡 trying to implement similar thing and have seen the light...

programmatix: Yes, in Java you'd usually create a class for this.  But Java 14 added Records, which can do this with minimal boilerplate, so I'd use those now.

modules/hello-world/examples/KvHelloWorldScoped.java
maria-robobug: Same here, I think I've forgotten to do it in these examples, 👍 I'll update.

modules/hello-world/examples/KvBulkHelloWorld.java
maria-robobug: Worth discussing in our docs sync meeting with PMs, for now will leave as is, but would be really nice to update this to use Records!

maria-robobug:
Start of the year seems like a good time for me to fix that flaky `Search.java` test...ignore the failure it's intermittent.

PR #296, opened 2021-12-08
Simplify how custom ports are specified when connecting to the cluster.

dnault:
Building SeedNodes is cumbersome, and no longer required in 3.2.

Update the docs to show the new way of specifying custom ports.

### DIFF
--- a/modules/howtos/examples/managing_connections.java
+++ b/modules/howtos/examples/managing_connections.java
@@ -14,21 +14,14 @@
  * limitations under the License.
  */
 
-import java.time.Duration;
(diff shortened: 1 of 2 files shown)


dnault:
What should I change to get the GitHub Action to pass? The code samples should compile, but should not be expected to execute (since they use custom ports which I assume are not actually functional).

maria-robobug:
@dnault that's a flaky test that I need to fix as part of tech debt, so ignore for now 👍 .

maria-robobug (approved):
LGTM!

PR #298, opened 2021-12-10
Doc 9216 import data (Java)

osfameron:
This is VERY draft, and quite horrific, have included dev team to ask for pointers on better more Java way of doing this.

* I realise this is syncronous, can rewrite to async, but not sure how best to verify that it works robustly / faster / etc.
* getting muddled with types (thus 2 functions for key/process based on type of Map)
* anything else horrible that needs fixing...

### DIFF
--- a/modules/howtos/examples/Import.java
+++ b/modules/howtos/examples/Import.java
@@ -47,7 +47,6 @@ import reactor.core.publisher.Flux;
 import java.util.stream.BaseStream;
 import java.util.HashMap;
 import java.time.Duration;
-import java.time.temporal.ChronoUnit;
 
 public class Import {
 
@@ -133,10 +132,9 @@ public class Import {
   
   // tag::importCSV[]
   public void importCSV() {
-    try {
-      CSVReaderHeaderAware csv = new CSVReaderHeaderAware(
-        new FileReader("howtos/examples/import.csv"));
-      
+    try (CSVReaderHeaderAware csv = new CSVReaderHeaderAware(
+        new FileReader("howtos/examples/import.csv"))) {
+            
       Map<String, String> row;
       while ((row = csv.readMap()) != null) {
         upsertRow(row);
@@ -173,9 +171,13 @@ public class Import {
           else { sink.next(row); }
           return state;
         }
-        catch (IOException e) { throw new Error("IOException"); }
-        catch (com.opencsv.exceptions.CsvValidationException e) { throw new Error("IOException"); }
-      });
+        catch (Exception e) { throw new RuntimeException(e); }
+      },
+      state -> {
+        try { state.close(); }
+        catch (Exception e) { throw new RuntimeException(e); }
+      }
+    );
 
     Flux<MutationResult> results = 
       rows
@@ -183,7 +185,7 @@ public class Import {
       .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
       
     results.subscribe(System.out::println);
-    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+    results.blockLast(Duration.ofSeconds(60));
     // tag::omit[]
     System.out.println("DONE");
     // end::omit[]
@@ -199,12 +201,11 @@ public class Import {
       .withIgnoreQuotations(true)
       .build();
     
-    try {
-      CSVReaderHeaderAware tsv =
+    try (CSVReaderHeaderAware tsv =
         new CSVReaderHeaderAwareBuilder(
           new FileReader("howtos/examples/import.tsv"))
         .withCSVParser(parser)
-        .build();
+        .build()) {
 
       Map<String, String> row;
       while ((row = tsv.readMap()) != null) {
@@ -260,8 +261,7 @@ public class Import {
           }
           return state;
         }
-        catch (IOException e) { throw new Error("IOException"); }
-        catch (com.opencsv.exceptions.CsvValidationException e) { throw new Error("IOException"); }
+        catch (Exception e) { throw new RuntimeException(e); }
       });
 
     Flux<MutationResult> results = 
@@ -270,7 +270,7 @@ public class Import {
       .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
       
     results.subscribe(System.out::println);
-    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+    results.blockLast(Duration.ofSeconds(60));
 
     System.out.println("DONE");
     // end::omit[]
@@ -318,7 +318,7 @@ public class Import {
           .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
           
       results.subscribe(System.out::println);
-      results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+      results.blockLast(Duration.ofSeconds(60));
     }
     // ...
     // tag::omit[]
@@ -335,10 +335,10 @@ public class Import {
   
   // tag::importJSONL[]
   public void importJSONL() {
-    try {
-        BufferedReader br =
+    try (BufferedReader br =
           new BufferedReader(
-            new FileReader("howtos/examples/import.jsonl"));
+            new FileReader("howtos/examples/import.jsonl"))) {
+              
         String line;
         while ((line = br.readLine()) != null) {
           Map<String,Object> row =
@@ -377,35 +377,35 @@ public class Import {
           .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
       
     results.subscribe(System.out::println);
-    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+    results.blockLast(Duration.ofSeconds(60));
     // tag::omit[]
     System.out.println("DONE");
     // end::omit[]
   }
   // end::importJSONL_batch[]
+
+  // tag::JsonDocument[]
+  class JsonDocument {
+    private final String id;
+    private final JsonObject content;
+
+    public JsonDocument(String id, JsonObject content) {
+      this.id = id;
+      this.content = content;
+    }
+
+    public String getId() {
+      return id;
+    }
+
+    public JsonObject getContent() {
+      return content;
+    }
+
+    @Override
+    public String toString() {
+      return "JsonDocument{id='" + id + "', content=" + content + "}";
+    }
+  }
+  // end::JsonDocument[]
 }
-
-// tag::JsonDocument[]
-class JsonDocument {
-  private final String id;
-  private final JsonObject content;
-
-  public JsonDocument(String id, JsonObject content) {
-    this.id = id;
-    this.content = content;
-  }
-
-  public String getId() {
-    return id;
-  }
-
-  public JsonObject getContent() {
-    return content;
-  }
-
-  @Override
-  public String toString() {
-    return "JsonDocument{id='" + id + "', content=" + content + "}";
-  }
-}
-// end::JsonDocument[]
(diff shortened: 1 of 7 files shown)


osfameron:
NB, though @mikereiche gave a workaround for the non-generic Map in Slack (thanks!), I haven't figured out how to fold it in without obscuring the intent of the code/docs.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,209 @@
...
+  //   JSON: Map<String,Object>
+  // and as generics annotations are lost at runtime we can't have
+  // a single overloaded method to deal with those.
+  //
+  // In real code, you might choose to cast from Object to Map<String,Object>
+  // or define multiple methods, but those confuse the example a little.
+  
+  // tag::upsertDocument[]
+  public void upsertRow(Map row) {
+
```
maria-robobug, line 117: Extra line here.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,314 @@
...
+      }
+      catch (java.io.FileNotFoundException e) {
+        System.out.println("handle FileNotFoundException...");
+      }
+      catch (java.io.IOException e) {
+        System.out.println("handle IOException...");
+      }
+      System.out.println("DONE");
+    }
+    // end::importJSON_batch[]
```
maria-robobug: Indentation looks a bit weird here.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,314 @@
...
+  
+    public JsonObject getContent() {
+      return content;
+    }
+  
+    @Override
+    public String toString() {
+      return "JsonDocument{id='" + id + "', content=" + content + "}";
+    }
+  }
```
maria-robobug, line 409: Think there's extra spacing here, should be the closing bracket so no space.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,314 @@
...
+          
+    docs.subscribe(System.out::println);
+    Flux<MutationResult> results = 
+      docs.flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
+      
+    results.subscribe(System.out::println);
+    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+
+    // ...
+    // tag::repeated[]
```
maria-robobug: Is there a section missing here, or should this tag be removed?

maria-robobug (approved):
👍 LGTM, just some minor comments!

modules/howtos/examples/Import.java
osfameron: ta! (copy/paste edit artefact, have tidied up in subsequent commit

modules/howtos/examples/Import.java
osfameron: believe I've removed this (renamed to ::omit also, which is more generally useful)

osfameron:
Updated with Reactor examples, @daschl @programmatix @dnault would really appreciate a once-over on this Java code as this is my first time use of Flux:: and various other libraries.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+    System.out.println(value);
+  }
+  // end::upsertDocument[]
+  
+  
+  // tag::importCSV[]
+  public void importCSV() {
+    try {
+      CSVReaderHeaderAware csv = new CSVReaderHeaderAware(
+        new FileReader("howtos/examples/import.csv"));
```
dnault: As a rule of thumb, code that creates a Reader or InputStream is also responsible for closing it. `try-with-resources` makes it easy:

    try (CSVReaderHeaderAware csv = new CSVReaderHeaderAware(
         new FileReader("howtos/examples/import.csv"))) {
    ...
    }

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+  // end::importCSV[]
+  
+  
+  // tag::importCSV_batch[]
+  public void importCSV_batch() {
+    // tag::omit[]
+    System.out.println("importCSV_batch");
+    // end::omit[]
+    
+    Flux<Map<String,String>> rows = Flux.generate(
```
dnault, line 162: Could use the Flux.generate() overload that takes a final cleanup callback to close the reader when the flux terminates. Just add a third argument: `state -> state.close()` (or if there's an IOException, catch and wrap in a RuntimeException)

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#generate-java.util.concurrent.Callable-java.util.function.BiFunction-java.util.function.Consumer-

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+        new FileReader("howtos/examples/import.csv")),
+      
+      (state, sink) -> {
+        try {
+          Map<String,String> row = state.readMap();
+          if (row == null) { sink.complete(); }
+          else { sink.next(row); }
+          return state;
+        }
+        catch (IOException e) { throw new Error("IOException"); }
```
dnault: Instead of catching different exceptions and re-throwing, how about signalling the error like:

    catch (Exception t) {
        sink.error(t);
    }

As an aside, `Error` is typically reserved for conditions that should never, *EVER* happen. Like, inconceivable situations that indicate something is fundamentally wrong with the code. If you just want to smuggle a checked exception, wrapping it in a `RuntimeException` is usually the way to go.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+        catch (com.opencsv.exceptions.CsvValidationException e) { throw new Error("IOException"); }
+      });
+
+    Flux<MutationResult> results = 
+      rows
+      .map(row -> preprocess(row))
+      .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
+      
+    results.subscribe(System.out::println);
+    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
```
dnault: NIT: `Duration.ofSeconds(60)`

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+        catch (IOException e) { throw new Error("IOException"); }
+        catch (com.opencsv.exceptions.CsvValidationException e) { throw new Error("IOException"); }
+      });
+
+    Flux<MutationResult> results = 
+      rows
+      .map(row -> preprocess(row))
+      .flatMap(doc -> reactiveCollection.upsert(doc.getId(), doc.getContent()));
+      
+    results.subscribe(System.out::println);
```
dnault: This code is upserting the documents twice, since `subscribe` and `blockLast` both subscribe to the flux (triggering its execution).

Recommendation is to remove the `subscribe`. If you want to print each mutation result, one way would be to chain `.doOnNext(System.out::println)` after `flatMap`.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+    // tag::omit[]
+    System.out.println("DONE");
+    // end::omit[]
+  }
+  // end::importCSV_batch[]
+
+  
+  // tag::importTSV[]
+  public void importTSV() {
```
dnault, line 197: NIT: There's an opportunity to refactor, since this is the same as the CSV version but with a differently configured CSV parser. Repeating the code is probably fine the documentation perspective, but we'll need to be careful to keep all of these versions in sync.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+  
+  // tag::importJSON_batch[]
+  public void importJSON_batch() {
+    // tag::omit[]
+    System.out.println("importJSON_batch");
+    // end::omit[]
+
+    try {
+      String content  = new String(
+        Files.readAllBytes(
```
dnault: NIT: Perhaps a note that this will fail if the whole file doesn't fit in memory? Ideally we'd want to use a JSON stream parser to read each object one by one, but maybe that's out of scope.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+  }
+  // end::importJSONL[]
+  
+  // tag::importJSONL_batch[]
+  public void importJSONL_batch() {
+    // tag::omit[]
+    System.out.println("importJSONL_batch");
+    // end::omit[]
+
+    Flux<String> lines = Flux.using(
```
dnault, line 368: NIT: Here's how Simon recommends doing this (see the "conclusion" section of https://simonbasle.github.io/2017/10/file-reading-in-reactor/ )

    Flux<String> lines = Flux.defer(() ->
            Flux.fromStream(Files.lines(thePath)));

dnault (changes requested):
These are some great examples.

modules/howtos/examples/Import.java
```diff
@@ -0,0 +1,411 @@
...
+    results.blockLast(Duration.of(60, ChronoUnit.SECONDS));
+    // tag::omit[]
+    System.out.println("DONE");
+    // end::omit[]
+  }
+  // end::importJSONL_batch[]
+}
+
+// tag::JsonDocument[]
+class JsonDocument {
```
maria-robobug: Hmm, I think the build is failing due to this class existing in https://github.com/couchbase/docs-sdk-java/blob/release/3.2/modules/hello-world/examples/KvBulkHelloWorld.java 

Because all these classes are part of the same "package", it's still accessible when using `class <Your-Class-Name-Here>`, so it comes up as a duplicate. 
I *think* `private class` could mitigate this, if you want to keep this in here. Or even rename it to something else.

Another option would be to reference the tag in `KvBulkHelloWorld` for the class, but it might look odd to the user if they don't see it when they access the full example on Github.

modules/howtos/examples/Import.java
osfameron: good idea! (I wasn't sure if I should name each exception to document which ones might happen? But happy to keep it simpler...)

modules/howtos/examples/Import.java
osfameron: oh, even better, this was in an example I, um, copy/pasted, and I did think `ChronoUnit` looked quite over-the-top

modules/howtos/examples/Import.java
osfameron: hmm, sink.error gives me a Generics type error, but I've used `throw new RuntimeException(e)` elsewhere at your suggestion anyway, so may keep with that for consistency.

modules/howtos/examples/Import.java
osfameron: woo, lemme give private a go

modules/howtos/examples/Import.java
osfameron: Thanks - he uses the `.using` construct in the main body of the article though, and the `.defer` only in the footnoted caveat.

Also .using does some error handling which the .defer incantation doesn't so not sure we'd gain any clarity by switching here?

modules/howtos/examples/Import.java
osfameron: Good catch. Yes, ideally would be streaming, using the Jackson Streaming API... 

I think this might be for a future version though (will check priority with @simon-dew) unless you happen to have time to contribute the code.

modules/howtos/examples/Import.java
osfameron: ah, I thought as it's push based, the code would be run once and the *result* sent to all subscribers?
I'll have to review my mental model... thanks, will look at .doOnNext!

modules/howtos/examples/Import.java
osfameron: Agreed, and yes, would like to refactor but risk of confusing docs intent.

dnault (approved):

modules/howtos/examples/Import.java
osfameron: (ended up nesting the class instead)

PR #299, opened 2021-12-20
Add up to date index mgmnt examples for dev-reorg

maria-robobug:
Docs issue: [DOC-9209](https://issues.couchbase.com/browse/DOC-9209)

### DIFF
--- a/modules/hello-world/examples/IndexHelloWorld.java
+++ b/modules/hello-world/examples/IndexHelloWorld.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2021 Couchbase, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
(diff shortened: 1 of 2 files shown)


modules/hello-world/examples/IndexHelloWorld.java
```diff
@@ -0,0 +1,126 @@
...
+import com.couchbase.client.java.manager.query.WatchQueryIndexesOptions;
+
+public class IndexHelloWorld {
+
+  public static void main(String... args) {
+    Cluster cluster = Cluster.connect("localhost", "Administrator", "password");
+
+    {
+      System.out.println("[primary]");
+      // tag::primary[]
```
simon-dew, line 30: I think the Java SDK lets you create and drop named primary indexes? (with the options object) Again, it's not strictly necessary, but it would be good to have examples of this also.

simon-dew (commented):
LGTM 👍

modules/hello-world/examples/IndexHelloWorld.java
maria-robobug: You can with create but not with drop for some reason from what I can see:
https://docs.couchbase.com/sdk-api/couchbase-java-client/com/couchbase/client/java/manager/query/DropPrimaryQueryIndexOptions.html#dropPrimaryQueryIndexOptions()

simon-dew: Yes of course, we already discussed this, my bad.

simon-dew:
Just to say @maria-robobug — I have used all the tagged examples in `IndexHelloWorld.java`, including the new `named-primary` tag. Thanks!

PR #302, opened 2022-01-18
[DOC-9543] - Document support for JDK 17

maria-robobug:
https://issues.couchbase.com/browse/DOC-9543

Updating to add support for JDK 17, let me know if anything is incorrect!

### DIFF
--- a/modules/project-docs/pages/compatibility.adoc
+++ b/modules/project-docs/pages/compatibility.adoc
@@ -8,10 +8,8 @@ Plus notes on Cloud, networks, and AWS Lambda.
 {description}
 
 The 3.2 SDK requires Java 8 or later to be installed, earlier versions will not work.
-_Java 11 is recommended_, which has various enhancements like lambda local variable type inference, profiling tools, and updated security features. 
+_Java 17 is recommended_, which has various enhancements like sealed classes, pattern matching for switch expressions (in preview), and further updates and improvements on core libraries.
 Most of the flavors available will do, although we may only provide support for OpenJDK and Oracle JDK going forward. 
-The Java SDK _has not been tested (and is not supported on) Java 12 or later_.
-
 
 == Couchbase Version/SDK Version Matrix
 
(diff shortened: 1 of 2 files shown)


modules/project-docs/pages/compatibility.adoc
```diff
@@ -8,10 +8,8 @@ Plus notes on Cloud, networks, and AWS Lambda.
 {description}
 
 The 3.2 SDK requires Java 8 or later to be installed, earlier versions will not work.
-_Java 11 is recommended_, which has various enhancements like lambda local variable type inference, profiling tools, and updated security features. 
+_Java 17 is recommended_, which has various enhancements like sealed classes, pattern matching for switch expressions (in preview), and further updates and improvements on core libraries.
```
RichardSmedley, line 11: I think here, or later, we need to say this is from 3.2.4 - unless @daschl is confident all 3.2.x will be fine on 17?

maria-robobug: Arun V mentioned from 3.2.4 in the ticket, so it does make sense to say from SDK 3.2.4 onwards  - would be good to get confirmation from @daschl though 👍 .

maria-robobug: @RichardSmedley have spoken to Michael, best to say from version `3.2.4` since that's  what has been tested, but in *theory* it should work with all of `3.2.x`.

I'll update this now.

modules/project-docs/pages/sdk-release-notes.adoc
```diff
@@ -22,6 +22,8 @@ At least Java 8 is required for current releases;
 see the xref:project-docs:compatibility.adoc#jdk-compat[Compatibility] section for details.
 We recommend running the latest LTS version (i.e. at the time of writing JDK 17) with the highest patch version available.
 Java 17 has various enhancements like sealed classes, pattern matching for switch expressions (in preview), and further updates and improvements on core libraries.
+
+NOTE: JDK 17 is only supported from SDK version 3.2.4 onwards.
```
RichardSmedley, line 26: Should we state what the fallback, latest-supported-version is, for 3.2.3.users?

RichardSmedley (approved):
:+1:

PR #300, opened 2022-01-14
Add section on Alpine Linux support

daschl:

PR #303, opened 2022-01-19
Alpine compat

RichardSmedley:
Fix for https://github.com/couchbase/docs-sdk-java/pull/300 going out of sync

### DIFF
--- a/modules/project-docs/pages/compatibility.adoc
+++ b/modules/project-docs/pages/compatibility.adoc
@@ -80,6 +80,36 @@ The following JDK releases are supported:
 Please make sure you run on one of the latest patch releases, since they provide stability improvements and security fixes in general.
 
 
+== OS Compatibility
+
+In general, the JVM removes worries about underlying OS compatibility, 
+and Couchbase JVM SDKs can be expected to run on all of the Operating Systems supported by xref:7.0@server:install:install-platforms.adoc[Couchbase Server].
+The JVM SDKs should also be expected to run on other commonly-available GNU/Linux distributions which support an appropriate JDK, but not all combinations can be tested -- notable exceptions are listed below.
+
+=== Alpine Linux Compatibility
+
+The https://netty.io[Netty] I/O library used by our JVM SDKs supports native optimizations to achieve higher throughput and lower latency 
+-- which the SDK enables by default. 
+Those modules are compiled against `glibc` and Alpine Linux uses `musl` instead. 
+It is not possible to determine if the SDK runs under Alpine Linux when it starts up, so SDK bootstrapping fails.
+
+Because Alpine Linux does not have full `glibc` support,
+we recommend that you use an alternative distribution if possible. 
+In case this is not possible, there are workarounds that can be applied.
+
+* Disable the native IO:
++
+[source,toml]
+====
+ClusterEnvironment env = ClusterEnvironment.builder().ioEnvironment(IoEnvironment.enableNativeIo(false)).build();
+====
+
+* Disable it in Netty itself via the following system property:
+`-Dcom.couchbase.client.core.deps.io.netty.transport.noNative=true`
+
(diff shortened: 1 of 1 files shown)


RichardSmedley:
Linking DOC-9555

modules/project-docs/pages/compatibility.adoc
```diff
@@ -80,6 +80,36 @@ The following JDK releases are supported:
 Please make sure you run on one of the latest patch releases, since they provide stability improvements and security fixes in general.
 
 
+== OS Compatibility
+
+In general, the JVM removes worries about underlying OS compatibility,
```
AV25242, line 85: Should we rephrase this sentence 
_JVM removes worries about to_ 
**JVM eliminates concerns about**

modules/project-docs/pages/compatibility.adoc
```diff
@@ -80,6 +80,36 @@ The following JDK releases are supported:
...
+In general, the JVM removes worries about underlying OS compatibility, 
+and Couchbase JVM SDKs can be expected to run on all of the Operating Systems supported by xref:7.0@server:install:install-platforms.adoc[Couchbase Server].
+The JVM SDKs should also be expected to run on other commonly-available GNU/Linux distributions which support an appropriate JDK, but not all combinations can be tested -- notable exceptions are listed below.
+
+=== Alpine Linux Compatibility
+
+The https://netty.io[Netty] I/O library used by our JVM SDKs supports native optimizations to achieve higher throughput and lower latency 
+-- which the SDK enables by default. 
+Those modules are compiled against `glibc` and Alpine Linux uses `musl` instead. 
+It is not possible to determine if the SDK runs under Alpine Linux when it starts up, so SDK bootstrapping fails.
```
AV25242: ### It is not possible to determine if the SDK runs under Alpine Linux when it starts up, so SDK bootstrapping fails.

I am lost here, what are we trying to say ? Are we saying  that had we know the OS we could have prevented bootstrapping failures ?

modules/project-docs/pages/compatibility.adoc
```diff
@@ -80,6 +80,36 @@ The following JDK releases are supported:
...
+
+Because Alpine Linux does not have full `glibc` support,
```
AV25242: ### Because Alpine Linux does not have full `glibc` support,

One of the previous lines mentioned use of `musl` as opposed to `glibc` but this line sounds like there is a `glibc` support but may be half baked ? again not clear if this line is needed or needs to be re-phrased ?

modules/project-docs/pages/compatibility.adoc
```diff
@@ -80,6 +80,36 @@ The following JDK releases are supported:
...
+we recommend that you use an alternative distribution if possible. 
+In case this is not possible, there are workarounds that can be applied.
+
```
AV25242, line 97: ### we recommend that you use an alternative distribution if possible. In case this is not possible, there are workarounds that can be applied.

Would something along these lines be incorporated ?

 **We highly recommend that you seek an alternative distribution where possible. If thats not a option then a possible work around that can be applied is as follows.**

modules/project-docs/pages/compatibility.adoc
AV25242: ### JVM removes worries about to

Worries sounds strong so trying to rephrase it something like the one below !

**JVM covers any concerns around underlying OS compatibility**

modules/project-docs/pages/compatibility.adoc
daschl: yeah, since then we could disable native IO by default. But we can't unfortunately.

modules/project-docs/pages/compatibility.adoc
daschl: no. alpine is musl only, and our native stuff needs glibc to work.

PR #305, opened 2022-01-28
DOC-9329: FLE Field Migration Tweak

osfameron:
Migrating a *field* from unencrypted is not related to migrating
from SDK 2 to SDK 3.

Moved this section as suggested by @dnault, and tweaked wording
very slightly to describe the kind of "migration" in the para.


dnault (approved):
LGTM. Thanks!

PR #308, opened 2022-02-11
Improve optimstic locking example

dnault:
Model a correct optimistic retry loop by getting the document inside the loop. If the replace failed due to CAS mismatch, read the document again to get a fresh CAS and see changes from the other actor.

### DIFF
--- a/modules/howtos/examples/KvOperations.java
+++ b/modules/howtos/examples/KvOperations.java
@@ -128,10 +128,11 @@ public class KvOperations {
       String id = "my-document";
       collection.upsert(id, JsonObject.create().put("initial", true));
 
-      GetResult found = collection.get(id);
-      JsonObject content = found.contentAsObject();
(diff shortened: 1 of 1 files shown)


dnault:
Looks like the "Test Code Samples" check failed due to:
```
com.couchbase.client.core.error.ServerOutOfMemoryException: The server is currently out of memory and cannot service the request
```

maria-robobug:
Well, Github hosted runners have about [7GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) of RAM, which should be plenty, wonder if I just need to allocate more memory to the KV service specifically?

Re-ran the build, and seems fine, the other failure is the flaky `Search.java` test which is a known issue! (**UPDATE**: now passed)
I'll keep an eye out if this memory error happens again in future though 👍

maria-robobug (approved):

PR #309, opened 2022-02-11
Gardening: Misc up KV docs improvements

dnault:
Use `indent=0` for all code snippets

Remove reference to deprecated `expiry()` method.

Document `preserveExpiry` and add an example.

Remove the expiry note partial because it doesn't apply to the Java SDK.


maria-robobug (approved):
LGTM!

PR #311, opened 2022-02-24
DOC-9752 - Further reading section should be java specific

maria-robobug:
https://issues.couchbase.com/browse/DOC-9752


osfameron (approved):
👍

PR #319, opened 2022-03-31
Updating transactions documentation for SDK integration in 3.3

programmatix:
Adding a migration guide for existing users of the
transactions library.

Moving pom.xml to the root.  This allows opening the full
project in an IDE, which is a better development workflow.

### DIFF
--- a/modules/howtos/examples/TransactionsMigration.java
+++ b/modules/howtos/examples/TransactionsMigration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Couchbase, Inc.
+ * Copyright (c) 2022 Couchbase, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/howtos/examples/TransactionsMigration.java
+++ b/modules/howtos/examples/TransactionsMigration.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2020 Couchbase, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// tag::imports[]
+
+import com.couchbase.client.core.msg.kv.DurabilityLevel;
+import com.couchbase.client.java.Bucket;
+import com.couchbase.client.java.Cluster;
+import com.couchbase.client.java.ClusterOptions;
+import com.couchbase.client.java.Collection;
+import com.couchbase.client.java.Scope;
+import com.couchbase.client.java.query.QueryResult;
+import com.couchbase.client.java.transactions.TransactionKeyspace;
+import com.couchbase.client.java.transactions.config.TransactionsCleanupConfig;
+import com.couchbase.client.java.transactions.config.TransactionsConfig;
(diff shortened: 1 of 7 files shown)


programmatix:
Ah the builds are failing due to me moving the pom.xml.  Is there a reason for the pom to be in modules though?  As mentioned it's a much better dev workflow if it's at the top-level.  I can then open docs-sdk-java as a project in the IDE, and have easy access to both the code samples (which get Intellisense and are compilable now), plus the adocs.

maria-robobug:
@programmatix it was mainly to keep everything in the same place for the test automation, but I can adjust the tests when I have a bit of time...

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -14,90 +14,57 @@ include::partial$acid-transactions-attributes.adoc[]
 
 include::{version-server}@sdk:shared:partial$acid-transactions.adoc[tag=intro]
 
-You may also want to start with our https://github.com/couchbaselabs/couchbase-transactions-java-examples[transactions examples repository],
-which features useful downloadable examples of using Distributed Transactions.
-
-Javadocs are available https://docs.couchbase.com/sdk-api/couchbase-transactions-java/index.html?overview-summary.html[online].
+// TODO update when available
```
maria-robobug, line 17: Not sure this will be necessary since it will be integrated in the SDK?

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -117,6 +84,15 @@ Those new to reactive programming may want to check out https://projectreactor.i
 
 include::{version-server}@sdk:shared:partial$acid-transactions.adoc[tag=lambda-ctx]
 
+The following helper logging methods will be used throughout the examples.
+Transaction logging is too verbose to be sent to the Java SDK event-bus, so instead it is stored in an in-memory log that can be accessed both from `TransactionResult` and `TransactionFailedException`.
+It is a good practice for the application to log any failed transactions into its own logging system, here represented with a `logger` object:
+[source,java]
```
maria-robobug, line 91:
```suggestion

[source,java]
```

modules/howtos/examples/TransactionsMigration.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright (c) 2020 Couchbase, Inc.
```
maria-robobug:
```suggestion
 * Copyright (c) 2022 Couchbase, Inc.
```

modules/howtos/pages/distributed-acid-transactions-migration-guide.adoc
```diff
@@ -0,0 +1,185 @@
+= Distributed Transactions Migration Guide
```
maria-robobug, line 1: I think this should probably move to the `Project Docs` section, under Compatibility.
https://docs-staging.couchbase.com/java-sdk/current/project-docs/compatibility.html 

But, I'm happy to keep this here for now in the interest of getting this on staging.
We can re-arrange where this goes in a separate PR.

maria-robobug (approved):
👍  LGTM, just some minor comments.
I'll look into fixing the `pom.xml` automation issue in a separate PR so we can get this on staging.

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
programmatix: Good point!

programmatix: Though, still it would be good to link to the transactions-specific portion of the javadocs from here, no?

modules/howtos/pages/distributed-acid-transactions-migration-guide.adoc
programmatix: Good idea, and sounds like a plan.

modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
maria-robobug: Yep, very true, let's link to that instead 👍

PR #320, opened 2022-04-13
transaction wording update for SDK API 3.3

osfameron:
see https://github.com/couchbase/docs-sdk-common/pull/49

### DIFF
--- a/modules/concept-docs/pages/durability-replication-failure-considerations.adoc
+++ b/modules/concept-docs/pages/durability-replication-failure-considerations.adoc
@@ -21,6 +21,7 @@ include::7.1@sdk:shared:partial$durability-replication-failure-considerations.ad
 include::7.1@sdk:shared:partial$durability-replication-failure-considerations.adoc[tag=performance]
 
 include::7.1@sdk:shared:partial$durability-replication-failure-considerations.adoc[tag=txns]
+* xref:project-docs:distributed-transactions-java-release-notes.adoc[Transactions Release Notes].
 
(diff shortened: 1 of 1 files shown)


modules/concept-docs/pages/durability-replication-failure-considerations.adoc
```diff
@@ -21,6 +21,7 @@ include::7.0@sdk:shared:partial$durability-replication-failure-considerations.ad
 include::7.0@sdk:shared:partial$durability-replication-failure-considerations.adoc[tag=performance]
 
 include::7.0@sdk:shared:partial$durability-replication-failure-considerations.adoc[tag=txns]
+* xref:project-docs:distributed-transactions-java-release-notes.adoc[Transactions Release Notes].
```
RichardSmedley, line 24: Is this in the right place?
Should it be in the Further Reading section?

osfameron: It was originally in the linked `tag=txns` partial, so this refactor leaves it in the same bulleted list.

osfameron: (intent is to be able to link the same partial from e.g. PHP sdk, without any Java specifics)

RichardSmedley: My digging into the changes without building got a little tangled  :)

maria-robobug: Makes sense to me, unless we think this wasn't the right place for it to begin with?

maria-robobug (approved):
👍

PR #323, opened 2022-04-26
3.3 release prep

osfameron:
* merge 3.2 branch work
* update Start Using page
* bump server to 7.1
* Update compatibility

### DIFF
--- a/modules/devguide/examples/CloudConnect.java
+++ b/modules/devguide/examples/CloudConnect.java
@@ -20,46 +20,28 @@ import com.couchbase.client.java.*;
 import com.couchbase.client.java.kv.*;
 import com.couchbase.client.java.json.*;
 import com.couchbase.client.java.query.*;
+import java.time.Duration;
 // end::imports[]
-
-// tag::capella-imports[]
-
-// Required for Capella connection configuration
-import com.couchbase.client.core.env.SecurityConfig;
-import com.couchbase.client.java.env.ClusterEnvironment;
-// end::capella-imports[]
 
 
 public class CloudConnect {
+  // tag::connect[]
   // Update these variables to point to your Couchbase Capella instance and credentials.
-  static String endpoint = "cb.njg8j7mwqnvwjqah.cloud.couchbase.com";
-  static String bucketname = "travel-sample";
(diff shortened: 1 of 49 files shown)


modules/devguide/examples/CloudConnect.java
```diff
@@ -14,97 +14,82 @@
...
+  // Update these variables to point to your Couchbase Capella instance and credentials.
+  static String endpoint = "cb.njg8j7mwqnvwjqah.cloud.couchbase.com";
+  static String bucketname = "travel-sample";
+  static String username = "username";
+  static String password = "Password!123";
+  
+  public static void main(String... args) {
+    // tag::connect[]
+    // Configure TLS
+    SecurityConfig.Builder securityConfig = SecurityConfig
```
maria-robobug: Do we still need this? Thought this was automatic now for Capella connections?

maria-robobug: Works for me in Go without a Security config, that's why I ask. 
Not sure if the same or different for Java 😕 .

osfameron: Don't know - I removed the Certificate wrangling, but I thought you still had to ensure it was TLS? @daschl to confirm?

osfameron: Ah ok. Lemme try without, even better if so...

osfameron: `An exception occured while executing the Java class. The DNS SRV record for 'cb.njg8j7mwqnvwjqah.cloud.couchbase.com' indicates TLS must be used when connecting to this cluster. Please enable TLS by setting the 'security.enableTLS' client setting to true. If the Cluster does not use a shared ClusterEnvironment, an alternative way to enable TLS is to prefix the connection string with "couchbases://" (note the final 's')`

osfameron: Yup, works great with `couchbases://` in connection string...
That's definitely the way to go for Hello World example, thanks!

osfameron: (also nice error message Michael & co!)

maria-robobug (approved):
👍

PR #331, opened 2022-05-11
Correction to Java SDK txns compatibility

dhaikney:
Included in the SDK since 3.3, rather than 3.1. (Needs validating but I believe this is correct)


RichardSmedley (approved):
LGTM
Thanks @dhaikney :+1:

PR #335, opened 2022-05-20
Bump version of Field-Level Encryption dependencies

dnault:
Also update the FLE examples to configure the ClusterEnvironment
using a lambda, so the user is not responsible for shutting down
the environment.


dnault:
Hi! Please let me know if I should take some action to get the checks to pass. Seems like it might be a baseline failure?

RichardSmedley:
> Hi! Please let me know if I should take some action to get the checks to pass. Seems like it might be a baseline failure?

Hi @dnault ,
It stems from the move of the pom.xml by transactions I think - tech debt that should be addressed soon.  :+1:

maria-robobug (approved):

PR #339, opened 2022-07-06
3.3.2 release

programmatix:


maria-robobug (approved):
LGTM, automation fails because the artifact is not published on Maven yet, but we can merge this now anyway.

PR #340, opened 2022-07-20
Restructure Distributed ACID Transactions documentation

maria-robobug:
Trying to make the documentation a bit easier to digest for users; I've split the doc into a howto guide and a concept page.

Page previews (including sdk-common changes, so it's easier to see it all together):
* [Transactions howto guide](https://maria-robobug.github.io/transactions-reorg/java-sdk/3.3/howtos/distributed-acid-transactions-from-the-sdk.html)

* [Transactions concept guide](https://maria-robobug.github.io/transactions-reorg/java-sdk/3.3/concept-docs/transactions.html)

NUX improvements:
* Splits howto and concept content as well as removing repetitive
  sections.

* Adds Capella in Prerequisites

* Removes repetitive examples from asciidoc where possible.


modules/howtos/pages/distributed-acid-transactions-from-the-sdk.adoc
```diff
@@ -1,50 +1,52 @@
...
+
+Couchbase Server::
++
+--
+* Couchbase Server (6.6.1 or above).
+* You should know how to perform xref:howtos:kv-operations.adoc[key-value] or xref:howtos:n1ql-queries-with-sdk.adoc[query] operations with the SDK.
+* Your application should have the relevant roles and permissions on the required buckets, scopes, and collections, to perform transactional operations.
+Refer to the xref:{version-server}@server:learn:security/roles.adoc[Roles] page for more details.
+* If your application is using xref:concept-docs:xattr.adoc[extended attributes (XATTRs)], you should avoid using the XATTR field `txn` -- this is reserved for Couchbase use.
+* NTP should be configured so nodes of the Couchbase cluster are in sync with time.
```
maria-robobug, line 44: Note that I added this **NTP** bit only for Couchbase Server, not entirely sure if you need to do this in Capella also?

@RichardSmedley do you have any idea?

RichardSmedley: AWS do provide NTP, and I'd be very surprised if we'd turned it off.
(But worth checking to be sure)

programmatix:
Some stuff in no particular order as I review:

- I love how the async and sync samples are presented now, with the tabs.
- I love how much more concise the new HOWTO is.
- I'm not 100% on parts of the concept docs (I know it's largely just the previously existing transactions mechanics section, but I wasn't 100% on that either).   Are we writing server docs here, and just doing it in the SDK docs because we've easier access to them?  But the concept docs are also merging in a part of SDK-specific HOWTO stuff, like how to configure cleanup.  More on this below.
- We have lost the 'here's all the imports' bit, which I do like, even though it is verbose.  I hate trying to get started with a new SDK with limited knowlege of its IDE, and having no idea where to find imports.  I guess it's replaced, sort of, by viewing examples on github (which don't work but I assume that's due to the staging).
- Maybe let's move the config section further down.  The defaults are fine for most users and it gets them into the action quicker.
- Splitting out some mini docs for niche functionality like tracing is smart.

And a bunch of stuff I initially thought was lost and then later found (mostly) in the concepts pages.  I've some concerns that the concepts pages aren't going to get read.  They're not super discoverable.  I think for any really crucial info, we might need to duplicate it in short form in the HOWTO, with a link to the more detailed concepts page, and/or ensure that it's in the server docs.

- We've lost the 'Single Query Transactions' section from the HOWTO.  Is this intentional?  Ah, found it on the concepts page.  Maybe that's ok - it's to a degree tertiary functionality.  Not certain.
- We've lost the 'real-world' MMORPG example - intentional?
- We've lost the 'Concurrency with Non-Transactional Writes' section, but this is a major real-world consideration for the app developer, who, let's face it, is often only to read the howto (if that).  I'm not sure we want to bury this away in concepts.
- We've lost any mention of cleanup.  Ah no it's in concept docs.  But there's not a clear distinction between HOWTO & concept on some of these pages, IMO, especially cleanup.  It's explaining what cleanup is - but also showing how to configure and monitor it.  I'm not certain if this is the best split.  But, I also don't want to overcomplicate the now nice-and-clean howto page..  Maybe cleanup should be a sub-page under the howto, like Tracing?  I'm not certain.
-- I think I feel the same about the single query transaction, custom metadata, and (to a lesser extent) error handling sections.  Again that mix of concept and howto.  
- We've lost mention of query requiring 7.0.  Ok I see it on the concepts page, maybe that's fine.
- "The cleanup process is detailed in [Error Handling]."  Should link to the dedicated cleanup doc, no?

Sorry, I know I'm raising a bunch of questions/concerns without having a proposed solution for all of them.  I _maybe_ feel like we should have more mini howtos, like Tracing, replacing some of the concept pages?  But that may also get messy.

maria-robobug:
I just want to say, I love all your feedback @programmatix! 

For the MMORPG example, I decided to remove it and focus on the example that had all the operations + some queries, though I am open to keeping it in if we have valid reasons (i.e. replacing the first example with it). I just want to avoid repetition where possible for the big examples.

> "The cleanup process is detailed in [Error Handling]." Should link to the dedicated cleanup doc, no?

Yah that's just a mistake 😄 , I'll clean that up.

I had internal struggle with what goes in concept and howto for sure, there *can* be some overlap, but I will have a discussion on the very good points you brought up with @RichardSmedley to see how we can address these.

maria-robobug:
Staged links should be updated soon...(process takes a few minutes, but they will be eventually consistent with the commits).

Having discussed with @RichardSmedley some of the key points, we decided to:

- Move concept docs under *Transactions* in the left hand navigation, so both concept and howto guide live in the same place. Grouped separately and in a hopefully obvious way to the user.
This should make it easier to find and maybe encourage devs to read the concepts also 😆 .

- Added back *Concurrency with Non-Transactional Writes*, think I missed that in my initial commit 👍 

- Added a mini howto for *Single Query Transactions*

- Fixed some mistakes

- We've decided to leave the Cleanup concept docs as is, it's always hard to draw the line between concept and howto. But we felt that since we've moved these docs under the same roof in *Transactions* it might be a little bit better.

> We've lost mention of query requiring 7.0. Ok I see it on the concepts page, maybe that's fine.

This felt more like something to include in a compatibility page even, so we decided to remove this from here.

maria-robobug:
FYI, if no objections by end of the week, I will merge and port the other SDKs to this format.

RichardSmedley (approved):
LGTM  :+1:

programmatix:
> For the MMORPG example, I decided to remove it and focus on the example that had all the operations + some queries, though I am open to keeping it in if we have valid reasons (i.e. replacing the first example with it). I just want to avoid repetition where possible for the big examples.

Ok, it does make sense, I'm just a bit sad to lose it as it's a nicer example than the first one - more real-world centric and less arbitrary bunch of operations.  But as you say, it was written pre-query.  At some point we probably want to develop a good real-world example.

I did a skim re-review and think it looks excellent now.  I think the new nav menu drop downs are a great way of managing the complexity and probably the best compromise possible.  As you say, pages like Cleanup are still something of a mix of concept and howto, but I don't that matters so much as discoverability.  Before, crucial howto stuff was hidden under the rather hidden Learn section.  Now we have a dollop of concept mixed into our tea of howto, but that feels a much lesser issue.

maria-robobug:
> Ok, it does make sense, I'm just a bit sad to lose it as it's a nicer example than the first one - more real-world centric and less arbitrary bunch of operations. But as you say, it was written pre-query. At some point we probably want to develop a good real-world example.

Definitely, I've kept the MMORPG example in the actual code file anyway, since it seems useful to keep for any future ideas/sample applications.

programmatix:
Btw I've requested on #transactions-dev that anyone interested have a look before this gets closed end of the week.

PR #342, opened 2022-08-03
Add units and fix typo on orphan logger

ingenthr:
Note that this same typo and lack of units is probably in the other SDK docs as well.

### DIFF
--- a/modules/howtos/pages/observability-orphan-logger.adoc
+++ b/modules/howtos/pages/observability-orphan-logger.adoc
@@ -72,10 +72,10 @@ The fields are kept compact so that the logs don't get too big, but since they a
 | Property       | Description
 | `b` | Name of the bucket
 | `r` | Remote hostname if dispatched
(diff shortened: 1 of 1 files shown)


dnault (approved):

maria-robobug (approved):
👍 Thanks both!

RichardSmedley:
> Note that this same typo and lack of units is probably in the other SDK docs as well.

I've had a quick check, @ingenthr , and .NET is the only one with a similar enough page structure to have that same table - and the error does not occur there.  :+1:

maria-robobug:
> I've had a quick check, @ingenthr , and .NET is the only one with a similar enough page structure to have that same table - and the error does not occur there.

Yep, I added the fix in .NET shortly after reviewing this PR: 
https://github.com/couchbase/docs-sdk-dotnet/commit/429e36acdb5838f990eb69c9e8de8b6bc122b555

PR #344, opened 2022-09-02
Removed timeoutConfig method from .builder()

sarahlwelton:
Per [the ticket](https://issues.couchbase.com/browse/DOC-8174), the `timeoutConfig` method doesn't exist under `builder()`. So, this PR is to remove that text from the example code.  

JIRA has this logged against 3.1, so I'd just want to know if we should propagate this to all 3.x versions. If so, should be an easy cherry pick across branches.

### DIFF
--- a/modules/project-docs/examples/Migrating.java
+++ b/modules/project-docs/examples/Migrating.java
@@ -66,7 +66,7 @@ public class Migrating {
       // tag::timeoutbuilder[]
       // SDK 3 equivalent
       ClusterEnvironment env = ClusterEnvironment.builder()
-          .timeoutConfig(TimeoutConfig.kvTimeout(Duration.ofSeconds(5))).build();
+          .build();
       // end::timeoutbuilder[]
     }
 


osfameron:
Note, best practice would be to quote the ticket number in the commit line like:

> `DOC-8174 Removed timeoutConfig method from .builder()`

modules/project-docs/examples/Migrating.java
```diff
@@ -66,7 +66,7 @@ public static void main(String... args) {
       // tag::timeoutbuilder[]
       // SDK 3 equivalent
       ClusterEnvironment env = ClusterEnvironment.builder()
-          .timeoutConfig(TimeoutConfig.kvTimeout(Duration.ofSeconds(5))).build();
+          .build();
```
osfameron, line 69: Is this correct?
Looks like this is now creating an empty cluster environment.

If this doesn't work, probably best to remove the whole section entirely.

That said https://docs.couchbase.com/java-sdk/current/ref/client-settings.html#config-builders does also suggest it exists?

sarahlwelton: Someone with more knowledge than me would have to comment... User that logged the ticket seemed to think it didn't exist so I thought it would be an easy fix! 

> There is no .timeoutConfig method behind the .builder() method...

osfameron: Hmm, don't see it in the API docs https://docs.couchbase.com/sdk-api/couchbase-java-client/com/couchbase/client/java/env/ClusterEnvironment.Builder.html

So perhaps this needs checking with SDK engineers and updating throughout docs?

sarahlwelton: Do you know who we could ask as far as developers go to verify?

PR #348, opened 2022-09-26
Gardening: Polish Cloud Connect example

dnault:
* Replace the call to Duration.parse with the more expressive Duration.ofSeconds

* Remove the copyright header from CloudConnect.java, since we don't want it to appear in the SDK Examples section of the Capella site.

* Add LICENSE (Apache License 2.0)


maria-robobug (approved):
👍 LGTM

PR #349, opened 2022-10-05
DOC-10368 - Configuration Profile Support

maria-robobug:
https://issues.couchbase.com/browse/DOC-10368

- Refactored the "Start Using" page documentation for [NUX Friendly docs](https://couchbasecloud.atlassian.net/browse/AV-36176) initiative. 
- Added an example for config profiles in "Start Using".
- Added Configuration Profiles documentation in the Client Settings page.

*Page Previews*
- [Start Using Page](https://maria-robobug.github.io/config-profiles/public/java-sdk/3.4/hello-world/start-using-sdk.html)
- [Client Settings Page](https://maria-robobug.github.io/config-profiles/public/java-sdk/3.4/ref/client-settings.html#constrained-network-environments)


modules/devguide/examples/StartUsingCloud.java
```diff
@@ -0,0 +1,85 @@
...
+            // Sets a pre-configured profile called "wan-development" to help avoid latency issues
+            // when accessing Capella from a different Wide Area Network
+            // or Availability Zone (e.g. your laptop).
+            .applyProfile("wan-development")
+            .build();
+
+    cluster = Cluster.connect(
+            connectionString,
+            ClusterOptions.clusterOptions(username, password).environment(env)
+    );
```
maria-robobug: @dnault or @mikereiche Please let me know if this is the correct way to set a configuration profile, is there anything else I should add?

maria-robobug:
FYI, examples will fail build/test checks because we can't test SDK 3.4 just yet - I have built the 3.4 snapshot locally though, so I have tested the examples there.

Once 3.4 is released then the tests and build checks should pass.

modules/devguide/examples/StartUsingCloud.java
maria-robobug: Also, will this still be marked as *Volatile* in SDK 3.4? 
If so, let me know and I'll add a note in the documentation.

maria-robobug: Looks like the *Volatile* question I had was answered in the SDK meeting - I will update the PR.

dnault: Hi Maria!

That code snippet would work, but would require the user to shut down the environment themselves, and would ignore the connection string scheme and parameters. In general, users should avoid the version of  `ClusterOptions.env` that takes a pre-built environment; instead they should the version that takes a callback to configure an environment builder, so the resulting ClusterEnvironment is owned and managed by the Cluster, and honors the connection string scheme and parameters.

Assuming `connectionString` starts with "couchbase**s**://", here's what I'd recommend:

```java
cluster = Cluster.connect(
    connectionString,
    ClusterOptions.clusterOptions(username, password).environment(env -> {
      // Sets a pre-configured profile called "wan-development" to help avoid latency issues
      // when accessing Capella from a different Wide Area Network
      // or Availability Zone (e.g. your laptop).
      env.applyProfile("wan-development");
    })
);
```

By the way, great explanation of when/why to use this profile!

Thanks,
David

dnault:
> FYI, examples will fail build/test checks because we can't test SDK 3.4 just yet - I have built the 3.4 snapshot locally though, so I have tested the examples there.
> 
> Once 3.4 is released then the tests and build checks should pass.

Java SDK snapshots are [usually...] published nightly to the Sonatype OSS snapshot repository. If you'd like to test against them, add this element as a child of the `<project>` element in your `pom.xml`

```xml
    <repositories>
        <repository>
            <id>sonatype-snapshots</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
            <releases><enabled>false</enabled></releases>
            <snapshots><enabled>true</enabled></snapshots>
        </repository>
    </repositories>

```
This should let you depend on snapshot versions, like `3.4.0-SNAPSHOT`.

PR #350, opened 2022-10-06
DOC-10496 - Document Platform Root CA usage

maria-robobug:


modules/howtos/pages/managing-connections.adoc
```diff
@@ -147,11 +148,30 @@ NOTE: Any TLS certificates must be set up at the point where the connections are
...
+--
+As of SDK 3.4, if your cluster's root certificate is signed by a well known CA (Certificate Authority) and is included in the JVM's trust store, it is no longer necessary to configure the certificate in the `securityConfig` settings.
 
-To configure encryption for the Java SDK:
+[source,java]
+----
+include::example$ManagingConnections.java[tag=tls-no-trust,indent=0]
+----
+
+However, you can still provide a certificate explicitly if necessary:
```
maria-robobug: @dnault from what I understand, well known CAs should be supported by default now, without any changes in the `securityConfig` settings.

However, I am unclear if this applies to Couchbase Server only, or Capella as well?

modules/howtos/pages/managing-connections.adoc
```diff
@@ -147,11 +148,30 @@ NOTE: Any TLS certificates must be set up at the point where the connections are
...
+--
+As of SDK 3.4, if your cluster's root certificate is signed by a well known Certificate Authority (e.g., GoDaddy, Verisign, etc) and is included in the JVM's trust store, you no longer need to configure the certificate in the `securityConfig` settings.
 
-To configure encryption for the Java SDK:
+[source,java]
+----
+include::example$ManagingConnections.java[tag=tls-no-trust,indent=0]
+----
+
+However, you can still provide a certificate explicitly if necessary:
```
maria-robobug: @dnault from what I understand, well known CAs should be supported by default now, without any changes in the securityConfig settings.

However, I am unclear if this applies to Couchbase Server only, or Capella as well?

modules/howtos/examples/ManagingConnections.java
```diff
@@ -166,6 +166,13 @@ public static void main(String... args) {
       // end::tls[]
     }
 
+    {
+      // tag::tls-no-trust[]
+      ClusterEnvironment env = ClusterEnvironment.builder()
```
dnault: Instead of showing users how to create their own ClusterEnvironment (which his something I'd personally like to discourage, since it means the user is responsible for shutting it down, and the environment won't honor the connection string scheme or parameters), would it make sense to show an example that uses a connection string that starts with "couchbases://" ?

modules/howtos/pages/managing-connections.adoc
```diff
@@ -147,11 +148,30 @@ NOTE: Any TLS certificates must be set up at the point where the connections are
...
++
+--
+The Java SDK includes Capella’s standard root certificate by default, which means you don’t need any additional configuration to enable TLS.
+Simply use `couchbases://` in your connection string, or `SecurityConfig.enableTls(true)` in your `SecurityConfig` settings.
+--
+
+Couchbase Server::
++
+--
+As of SDK 3.4, if your cluster's root certificate is signed by a well known Certificate Authority (e.g., GoDaddy, Verisign, etc) and is included in the JVM's trust store, you no longer need to configure the certificate in the `securityConfig` settings.
```
dnault: > and is included in the JVM's trust store,

In the case we're talking about here, the cluster's root certificate doesn't need to be in the JVM's trust store. The cluster's root certificate just needs to be *issued* (signed) by a CA whose certificate is in the JVM's trust store.

Another detail that might be useful: when we talking about the JVM's trust store, we're specifically referring to the `cacerts` file.

modules/howtos/pages/managing-connections.adoc
dnault: Hi Maria. Capella's root certificate is not signed by a well-known CA. However, starting with SDK 3.3.0, we bundle the Capella CA certificate with the SDK and trust it by default (unless the user specifies different certificates to trust).

The change in 3.4 is that in addition to trusting the Capella CA certificate by default, we *also* trust all the CA certificates in the JVM's trust store (the `cacerts` file).

The upshot:
* If you're connecting to Capella, you don't need to configure certificates.
* If you're connecting to a Couchbase Server cluster whose root certificate was issued (signed) by a CA whose certificate is in the JVM's `cacerts` file (this includes well-known CAs, plus any CA certificates the user adds themselves), you don't need to configure certificates.

modules/howtos/pages/managing-connections.adoc
```diff
@@ -161,7 +181,7 @@ It is important to make sure you are transferring the certificate in an encrypte
 If you are running on `localhost` and just want to enable TLS for a development machine, just copying and pasting it suffices
 -- _so long as you use `127.0.0.1` rather than `localhost` in the connection string_.
 This is because the certificate will not match the name _localhost_.
-Setting `TLSSkipVerify` is a workaround if you need to use ` couchbases://localhost`.
+Setting `TLSSkipVerify` is a workaround if you need to use `couchbases://localhost`.
```
dnault: I can't find any reference to `TLSSkipVerify` in the Java SDK. I suspect the corresponding client might be `security.enableHostnameVerification` which you can set to `false`, either in the connection string or configuring the cluster environment's SecurityConfig`.

I was going to cite a documentation reference, but it looks like this setting is undocumented? Here's where I looked for it: https://docs.couchbase.com/java-sdk/current/ref/client-settings.html#security-options

Or... if the intent is to completely disable all TLS verification and trust any certificate, the way to do that in Java today is to configure the environment'd SecurityConfig to use `InsecureTrustManagerFactory.INSTANCE`

PR #356, opened 2022-11-03
Mention Couchbase SDK 2 Migration Kit

dnault:
The migration kit is a home for bits of code that might help users migrating from SDK 2 to SDK 3. It currently has a version of the experimental Query DSL from SDK 2, modified to work with SDK 3. It also has bridge classes for migrating subdoc requests.


ingenthr:
Seems good by me.  Minor concern is there were concerns about how things are worded in the documentation to be cautious about implied support for those things (though you worded the README in the repo quite well).  I'll leave that concern up to @RichardSmedley.

osfameron (approved):
Agree with @ingenthr that the wording looks good.
As Richard is out, I'll approve to unblock if you want to merge now, and review later.

RichardSmedley (approved):
LGTM

PR #361, opened 2023-01-30
Tidying the OpenTelemtry documentation

programmatix:
The Jaeger example is outdated and has bitrotted due to the OpenTelemetry APIs changing.

But there are dozens of OpenTelemetry backends.  Rather than providing instructions on connection to just one, let's instead standardise on one consistent and
simple approach, which is exporting OpenTelemetry traces in OTLP format.  These can be sent to any OTLP-compatible backend, which is pretty much all of them, including Jaeger.

Also added some troubleshooting tips.


RichardSmedley (approved):
:+1:

PR #363, opened 2023-02-16
Polish "Getting Started"

dnault:
* Rename StartUsingCloud -> StartUsingCapella

* Make it clear that the scheme ("couchbases://") is part of the connection string.

* Most users will customize the cluster environment, so make that the primary example.

* Relegate the simple version of Cluster.connect() to a TIP.

* Use `127.0.0.1` instead of `localhost` because the default self-signed certificates are issued for `127.0.0.1`, and fail with `localhost`.

### DIFF
--- a/modules/devguide/examples/Cloud.java
+++ b/modules/devguide/examples/Cloud.java
@@ -42,7 +42,7 @@ import com.couchbase.client.java.query.QueryResult;
 
 /**
  * Example of Cas (Check and Set) handling in Java for the Couchbase Developer
- * Guide ported from 2.x. See StartUsingCloud.java for 3.x.
+ * Guide ported from 2.x. See StartUsingCapella.java for 3.x.
  */
 
 public class Cloud {
--- a/modules/devguide/examples/StartUsingCapella.java
+++ b/modules/devguide/examples/StartUsingCapella.java
@@ -8,33 +8,25 @@ import com.couchbase.client.java.query.*;
 import java.time.Duration;
 // end::imports[]
 
-
-public class StartUsingCloud {
+public class StartUsingCapella {
   // tag::connect-info[]
   // Update these variables to point to your Couchbase Capella instance and credentials.
-  static String connectionString = "cb.<your-endpoint-here>.cloud.couchbase.com";
+  static String connectionString = "couchbases://cb.<your-endpoint-here>.cloud.couchbase.com";
   static String username = "username";
   static String password = "Password!123";
   static String bucketName = "travel-sample";
   // end::connect-info[]
 
   public static void main(String... args) {
-    // tag::connect-string[]
-    // Simple connection.
-    Cluster cluster = Cluster.connect("couchbases://" + connectionString, username, password);
-    // end::connect-string[]
-    cluster.disconnect();
-
     // tag::connect-env[]
-    // Custom environment connection.
-    cluster = Cluster.connect(
-            "couchbases://" + connectionString,
-            ClusterOptions.clusterOptions(username, password).environment(env -> {
-              // Sets a pre-configured profile called "wan-development" to help avoid latency issues
-              // when accessing Capella from a different Wide Area Network
-              // or Availability Zone (e.g. your laptop).
-              env.applyProfile("wan-development");
-            })
+    Cluster cluster = Cluster.connect(
+        connectionString,
+        ClusterOptions.clusterOptions(username, password).environment(env -> {
+          // Sets a pre-configured profile called "wan-development" to help avoid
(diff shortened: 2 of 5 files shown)


modules/devguide/examples/Cloud.java
```diff
@@ -42,7 +42,7 @@
 
 /**
  * Example of Cas (Check and Set) handling in Java for the Couchbase Developer
- * Guide ported from 2.x. See StartUsingCloud.java for 3.x.
+ * Guide ported from 2.x. See StartUsingCapella.java for 3.x.
  */
 
 public class Cloud {
```
AV25242, line 48: Why is this class alone still called Cloud and not Capella :)

AV25242: And the file name ;)

dnault: Hi Arun! I don't know. I didn't want to touch it, because it wasn't related to the change at hand... and I needed to resist the temptation to fix everything at once ;-)

dnault:
Hi @RichardSmedley , should I be worried about the failed "Test Code Samples" check, or is it a baseline failure?

modules/devguide/examples/Cloud.java
AV25242: No that makes perfect sense. the only other recommendation I had was you mentioned `address` to be precise probably call it hostaddress because thats what it is ?

RichardSmedley:
> Hi @RichardSmedley , should I be worried about the failed "Test Code Samples" check, or is it a baseline failure?

Baseline failure - sth to do with the docker image we "temporarily" changed to for $REASONS iirc, and not fixed before we lost Maria.
It's top of the todo list for whoever lands here next  ;)

RichardSmedley:
"Polish Getting Started" - I thought you meant _Przewodnik wprowadzający_ :-) 
Translations are still some years off ;)

modules/hello-world/pages/start-using-sdk.adoc
```diff
@@ -230,11 +221,17 @@ include::example$StartUsing.java[tags=connect-env,indent=0]
 --
 ====
 
-Following successful authentication, add this code snippet to access your `Bucket`:
+TIP: There's also a simpler version of `Cluster.connect()` for when you don't need to customize the cluster environment:
```
RichardSmedley: I think you want the snippet in the TIP?
In Antora (well, asciidoc) you can do that with:

```
[TIP]
.Simpler Connection
====
There's also a simpler version of `Cluster.connect()` for when you don't need to customize the cluster environment:

[source,java]
----
include::devguide:example$SimpleConnect.java[tags=connect-string,indent=0]
----
====
```

modules/devguide/examples/StartUsingCapella.java
```diff
@@ -8,33 +8,27 @@
 import java.time.Duration;
 // end::imports[]
 
-
-public class StartUsingCloud {
+public class StartUsingCapella {
   // tag::connect-info[]
   // Update these variables to point to your Couchbase Capella instance and credentials.
-  static String connectionString = "cb.<your-endpoint-here>.cloud.couchbase.com";
+  static String address = "cb.<your-endpoint-here>.cloud.couchbase.com";
```
raycardillo: After thinking myself in circles on this one, I do think `connectionString` is probably still best here, or refactor a bit more (per my next comments). Technically the full connection string includes the scheme, so I get what you meant here, but maybe we just make this the full connection string instead? I'm thinking from the sense that it's not really a "hostname" or "address" either but rather our "connection string format" as documented throughout (e.g., https://docs.couchbase.com/java-sdk/current/howtos/managing-connections.html#connection-strings). Our other samples are a bit hit and miss in this regard as well, but several just put the whole string in (or don't even have a variable) and use the full "connection string" here.

raycardillo (changes requested):
I only had one additional comment to add other than what @RichardSmedley already raised. Looks so much better though overall. Thanks for doing this!

modules/hello-world/pages/start-using-sdk.adoc
dnault: Thanks! Done.

modules/devguide/examples/StartUsingCapella.java
dnault: Done!

modules/devguide/examples/Cloud.java
dnault: Resolving this conversation, since there's no longer a separate variable for the address.

RichardSmedley (approved):
Thanks @dnault - this LGTM  :+1:

PR #364, opened 2023-03-06
Test framework expansion

tom-rosewell:
Added a new function in the Makefile that bypasses the tests to just run the given code sample in the container.
This allows the developer to see the code output.


tom-rosewell:
I also moved the Dockerfile out of the modules/test directory because it didn't really make any sense to be in there

tom-rosewell:
AddEnrollments.java passed on my local env. I will look into it

tom-rosewell:
More work needed

PR #369, opened 2023-03-16
Add colour to the command output when the server is being configured

tom-rosewell:


tom-rosewell:
resolving local git commit discrepancy

PR #395, opened 2023-11-29
DOC-11688 Update logging docs

dnault:
Motivation
----------
It's all about SLF4J now!

Modifications
-------------
Rewrote most of the "Logging" section to reflect how SDK 3.5 does logging.

Updated the Log4j examples to use Log4j 2.

Removed Logback examples. Logback is compatible, but we don't need to document how to configure _every_ SLF4J binding.

Added a section about using the SLF4J binding for java.util.logging (JUL), since this is a migration path for people who were not previously using SLF4J.

Removed the section on the SDK's logger config, since those settings are deprecated in favor of configuring everything via the SLF4J binding.

Reformatted the code example for configuring java.util.logging (JUL) log level.

### DIFF
--- a/modules/howtos/examples/CollectingInformationAndLogging.java
+++ b/modules/howtos/examples/CollectingInformationAndLogging.java
@@ -51,28 +51,17 @@ public class CollectingInformationAndLogging {
     collection = scope.collection("airport");
     // end::connection_1[]
   }
+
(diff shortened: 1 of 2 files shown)


RichardSmedley (approved):
LGTM 
Thanks @dnault   :+1:

PR #401, opened 2024-01-19
Document vector search

programmatix:
My understanding is it's fine for this to be merged now, but the docs site should not be built from it until later.


RichardSmedley (approved):
Thanks @programmatix - will merge into this branch too, as basis for the reworking of the page...

PR #402, opened 2024-01-19
Document vector search

programmatix:


RichardSmedley (approved):
Thanks @programmatix - LGTM :+1:

PR #405, opened 2024-02-26
JCBC-2124 range scan and subdoc read from replica

dnault:
and incidentally, use `List.of` instead of `Collections.singletonList` or `Arrays.asList`


RichardSmedley (approved):
Thanks @dnault - LGTM :+1:

PR #419, opened 2024-08-29
[AV-82442] Update Java release 3.5 docs for Free Tier

Rebecca-Martinez007:
Removed mentions of trial in the Java SDK docs. Only locations found were the start-using-sdk.adoc page and the n1ql-queries-with-sdk.adoc. Used git cherry-pick from 3.4 to complete.


RichardSmedley (approved):
:+1:

PR #428, opened 2024-11-20
Add item missing from 3.7.4 relnotes

programmatix:


modules/project-docs/pages/sdk-release-notes.adoc
```diff
@@ -125,6 +125,9 @@ This has now been fixed.
 
 ==== Improvements
 
+* https://jira.issues.couchbase.com/browse/JCBC-2147[JCBC-2147]:
+New APIs added to allow getting KV documents from a preferred server group.
+This can support use-cases such as reading from replicas that are in the same availability zone as the application.
```
RichardSmedley:
```suggestion
New APIs added to allow getting KV documents from a preferred server group.
This feature allows the implementation of network optimization when traffic cost between server groups is higher than in the local group.
In this case the application might select preferred server group in the connection options, and later opt-in for local operations during replica reads. 
```

RichardSmedley (commented):
Added some expanded wording as a suggestion.

PR #430, opened 2024-11-27
Update client setting examples to use config callback.

dnault:
Motivation
----------
Config callback has advantages over manually creating your own ClusterEnvironment:

* Enables using connection string parameters and Java system properties to configure client settings.

* It's harder for the user to accidentally clobber client settings set by other parts of their code.

* The SDK automatically shuts down the environment for you, eliminating a potential resource leak.

Modifications
-------------
Update code snippets to use config callback.

Add an anchor for each client setting.

Use tables to structure client setting info.

Document connection string parameter names.

Fix `io.idleHttpConnectionTimeout` default value (a while ago it was reduced from 4.5s to 1s)

Warn that manually-created ClusterEnvironments are incompatible with connection string parameters and Java system properties.

Document that `couchbases://` overrides the `security.enableTls` setting.

### DIFF
--- a/modules/howtos/examples/ManagingConnections.java
+++ b/modules/howtos/examples/ManagingConnections.java
@@ -14,24 +14,29 @@
  * limitations under the License.
  */
 
+import com.couchbase.client.core.env.IoConfig;
+import com.couchbase.client.core.env.SecurityConfig;
+import com.couchbase.client.java.*;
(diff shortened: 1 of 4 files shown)


RichardSmedley (approved):
Thanks @dnault - great to have all of this.  :+1:

PR #431, opened 2024-12-05
Attempt to fix client settings table formatting

dnault:
Motivation
========
Fix this formatting gore from 8f87aeb3e254f7e6ed8421264d193b2c2d5d7698

table-header-formatting-gore_720

Modifications
=============
Remove non-breaking spaces; better to break at a space than to hyphenate.

Use autowidth; see if that's better than minimizing the width of the first column.


RichardSmedley (approved):
LGTM :+1:

PR #433, opened 2024-12-10
Add sub-heading for each client setting

dnault:
Motivation
----------
Make it easier to link to a specific section


RichardSmedley (approved):
I'm not entirely sure about losing the block formatting -
but I'm willing to put this through and live with it for a while,
and see how it goes.  :+1:

PR #448, opened 2025-05-28
Update snapshot repository info

dnault:
Motivation
----------
Now that we've migrated from OSSRH to Central Portal, snapshots are in a different location.


RichardSmedley (approved):
LGTM :+1:

PR #449, opened 2025-06-10
Fix typo in download link for the 3.8.1 release

yeikel:
Following the pattern of all other entries, it seems to me that this is an unintentional typo


yeikel:
@dnault This one may add value. But it is unclear if you'd rather handle it internally like #450

dnault (approved):
Wonderful, thank you!

PR #467, opened 2025-10-16
Fix otel.yaml config file in otel metrics exaple

DemetrisChr:
By default the receivers listen to `localhost`, so an SDK outside the docker container is unable to reach it. For this example to work we should specify `0.0.0.0`. Judging from the docs, this sounds like a recent change: https://opentelemetry.io/docs/collector/configuration/#basics


RichardSmedley (approved):
Thanks @DemetrisChr - LGTM :+1:

PR #488, opened 2026-04-28
Release notes for 3.5.4 and 3.6.4

dnault:


ingenthr (approved):
LGTM.  This does seem to set a new precedent with community member contributions though.  I don't think we've called that out in the past in the RN.

RichardSmedley (approved):
👍

RichardSmedley:
> LGTM. This does seem to set a new precedent with community member contributions though. I don't think we've called that out in the past in the RN.

No, we've pushed back against that in the past.
Letting this one through, given the circs.

PR #494, opened 2026-08-03
Clarify which JDKs we test and support

dnault:
Move the Oracle JDKs to a separate list and call out that they receive less testing.

Update the SDKMAN! example so it always installs the latest stable version of Eclipse Temurin instead of an obsolete version of the Oracle JDK.

### DIFF
--- a/modules/project-docs/pages/compatibility.adoc
+++ b/modules/project-docs/pages/compatibility.adoc
@@ -30,7 +30,7 @@ The following OpenJDK releases are supported:
 * https://adoptium.net/[OpenJDK 11 with HotSpot JVM]
 * https://adoptium.net/[OpenJDK 8 with HotSpot JVM]
 
-The following Oracle JDK releases are supported but do not receive the same level of testing due to Oracle license issues:
+The following Oracle JDK releases are also supported but we do not do any testing with them owing to Oracle license issues:
 
 * https://www.oracle.com/java/technologies/downloads/#java17[Oracle JDK 17]
 * https://www.oracle.com/java/technologies/downloads/#java11[Oracle JDK 11]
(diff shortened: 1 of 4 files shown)


RichardSmedley (dismissed):
LGTM 👍

modules/project-docs/pages/compatibility.adoc
```diff
@@ -20,22 +20,26 @@ _Java {java_latest_lts_version} is recommended_.
...
+The following OpenJDK releases are supported:
 
 * https://adoptium.net/[OpenJDK 25 with HotSpot JVM] (recommended)
 * https://adoptium.net/[OpenJDK 21 with HotSpot JVM]
 * https://adoptium.net/[OpenJDK 17 with HotSpot JVM]
-* https://www.oracle.com/java/technologies/downloads/#jdk17[Oracle JDK 17]
-* https://adoptium.net/[OpenJDK 11] (Hotspot recommended) or https://www.oracle.com/java/technologies/downloads/#jdk11[Oracle JDK 11]
-* https://adoptium.net/[OpenJDK 1.8 with HotSpot JVM]
-* https://www.oracle.com/java/technologies/downloads/#java8[Oracle JDK 1.8]
+* https://adoptium.net/[OpenJDK 11] (Hotspot recommended)
```
raycardillo: After our discussion, this line should just look like the others:
`OpenJDK 11 with HotSpot JVM`

modules/project-docs/pages/compatibility.adoc
```diff
@@ -20,22 +20,26 @@ _Java {java_latest_lts_version} is recommended_.
...
+* https://adoptium.net/[OpenJDK 8 with HotSpot JVM]
 
-Please make sure you run on one of the latest patch releases, since they provide stability improvements and security fixes in general.
+The following Oracle JDK releases are supported but do not receive the same level of testing due to Oracle license issues:
```
raycardillo: We don't test with them at all, so something more like "but we do not do any testing with them due to Oracle license issues"

raycardillo (changes requested):
Just a few comments.

RichardSmedley (approved):
LGTM 👍 

DOC-14563

